PR #259, opened 2025-11-14
Template detection updates

kddejong:
*Issue #, if available:*
- Template detection updates
*Description of changes:*

### DIFF
--- a/src/services/guard/GuardService.ts
+++ b/src/services/guard/GuardService.ts
@@ -178,7 +178,7 @@ export class GuardService implements SettingsConfigurable, Closeable {
             fileType === CloudFormationFileType.Unknown ||
             fileType === CloudFormationFileType.Empty
         ) {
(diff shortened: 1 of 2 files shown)


src/services/guard/GuardService.ts
```diff
@@ -172,8 +172,13 @@ export class GuardService implements SettingsConfigurable, Closeable {
...
 
-        if (!fileType || fileType === CloudFormationFileType.Unknown) {
-            this.telemetry.count(`validate.file.${CloudFormationFileType.Unknown}`, 1);
+        if (
+            !fileType ||
+            fileType === CloudFormationFileType.Other ||
+            fileType === CloudFormationFileType.Unknown ||
+            fileType === CloudFormationFileType.Empty
+        ) {
+            this.telemetry.count(`validate.file.${fileType}`, 1);
```
Zee2413: What is this metric name in the case `!fileType`? Is that what we want?

PR #261, opened 2025-11-14
Create databases on start, increase database size and improve logging…

satyakigh:
## Summary

This PR refactors the CloudFormation Language Server's data storage layer with a focus on initialization, performance, and observability improvements.

### 1. Database Initialization Strategy
• **Before**: Databases were created lazily on first access via getOrCreate()
• **After**: Databases are created upfront during LMDB initialization
• Stores are now explicitly defined using a StoreName enum instead of arbitrary strings

### 2. Increased Database Size
• LMDB max size increased from 100MB → 250MB 
  • With SAM schemas the total database is already > 100MB
• Added configuration for page size (8192) and remapping chunks for better performance

### 3. Logging & Metrics
• Added system information logging on startup (OS platform, architecture, Node.js version, V8 version)
• Improved log messages to be more concise and informative:
  • "X public schemas retrieved" instead of "X resource schemas retrieved"
  • "Combined schemas from public=X, private=Y, SAM=Z"
• Enhanced LMDB metrics with per-store size and entry counts
• Added global database usage percentage metric

### **LMDB Memory Access**

Before: Databases created inside transactionSync() with race conditions
After: All databases created synchronously during constructor before any async operations
typescript

Impact:
• **No lazy async database creation** - eliminates race conditions where multiple threads try to create same store
• **Memory mapped before use** - all LMDB memory regions mapped at startup, no runtime mapping failures
• **remapChunks: true** - allows safe remapping if database grows beyond initial allocation
• **Predictable memory layout** - all stores exist in memory map from start, no fragmentation from lazy creation

### **Async/Sync Safety**

Fix: Removed void promises that silently swallow errors
Impact:
• **No silent failures** - async errors in cache operations now visible
• **Proper error propagation** - LMDB write failures won't be ignored
• **Safer concurrent access** - errors from race conditions will be logged instead of lost

### **Memory Unmapping Risk**

Before: Stores could be accessed before memory mapping completed (lazy creation)

After:
• All memory regions mapped synchronously at startup
• No dynamic store creation means no runtime unmapping/remapping during access

### DIFF
--- a/src/datastore/LMDB.ts
+++ b/src/datastore/LMDB.ts
@@ -6,7 +6,7 @@ import { ScopedTelemetry } from '../telemetry/ScopedTelemetry';
 import { Telemetry } from '../telemetry/TelemetryDecorator';
 import { TelemetryService } from '../telemetry/TelemetryService';
 import { pathToArtifact } from '../utils/ArtifactsDir';
-import { DataStore, DataStoreFactory } from './DataStore';
+import { DataStore, DataStoreFactory, StoreName } from './DataStore';
 import { encryptionStrategy } from './lmdb/Utils';
 
 const log = LoggerFactory.getLogger('LMDB');
@@ -15,7 +15,7 @@ export class LMDBStore implements DataStore {
     private readonly telemetry: ScopedTelemetry;
 
     constructor(
-        private readonly name: string,
+        public readonly name: StoreName,
         private readonly store: Database<unknown, string>,
     ) {
         this.telemetry = TelemetryService.instance.get(`LMDB.${name}`);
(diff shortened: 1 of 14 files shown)


src/datastore/LMDB.ts
```diff
@@ -77,44 +88,22 @@ export class LMDBStoreFactory implements DataStoreFactory {
             },
             2 * 60 * 1000,
         );
+
+        log.info(`Initialized LMDB ${Version} at ${rootDir}`);
```
Zee2413: Do we want to avoid logging rootDir as it has absolute path? If customer shares log file when reporting crash, it will expose the full path.

Zee2413 (approved):

kddejong (approved):

PR #268, opened 2025-11-15
Better cancellation token error suppression

kddejong:
*Issue #, if available:*

*Description of changes:*


src/utils/Delayer.ts
```diff
@@ -1,3 +1,10 @@
+export class CancellationError extends Error {
```
satyakigh: I was trying something like this as well https://github.com/aws-cloudformation/cloudformation-languageserver/pull/264

PR #286, opened 2025-11-18
Remove override/V2 SDK

atennak1:

### DIFF
--- a/assets/featureFlag/prod.json
+++ b/assets/featureFlag/prod.json
@@ -6,7 +6,39 @@
       "enabled": false
     },
     "EnhancedDryRun": {
-      "enabled": true
+      "enabled": true,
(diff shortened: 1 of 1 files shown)


src/services/CfnService.ts
```diff
@@ -193,6 +195,40 @@ export class CfnService {
...
+                    new DescribeEventsCommand({
+                        ...params,
+                        NextToken: nextToken,
+                    }),
+                )) as unknown as DescribeEventsCommandOutput;
+
+                if (result) {
+                    operationEvents.push(...(response.OperationEvents ?? []));
+                } else {
+                    result = response;
```
rongyamazon: nit: only this assignment need to be in the if statement

rongyamazon (approved):

PR #299, opened 2025-11-23
Remove online status popup

kddejong:
*Issue #, if available:*

*Description of changes:*
- Remove online status popup


Zee2413 (commented):
We can delete the online status class all together. We don't need a log on server side. Likely we will never go looking for it. All online feature failures due to lack of network will be logged by the online guard.

PR #316, opened 2025-11-25
Reduce standard waiter times

kddejong:
*Issue #, if available:*

*Description of changes:*
- Reduce standard waiter times

### DIFF
--- a/src/services/CfnService.ts
+++ b/src/services/CfnService.ts
@@ -60,16 +60,15 @@ import {
     ChangeSetType,
 } from '@aws-sdk/client-cloudformation';
 import { WaiterConfiguration, WaiterResult } from '@smithy/util-waiter';
-import { SettingsManager } from '../settings/SettingsManager';
+import { AwsClientSettings, DefaultSettings } from '../settings/Settings';
 import { DeploymentMode } from '../stacks/actions/StackActionRequestType';
 import { Count } from '../telemetry/TelemetryDecorator';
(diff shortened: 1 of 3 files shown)


tools/generate-release-manifest.ts
```diff
@@ -91,6 +91,7 @@ function generateManifest() {
 
     for (const env of ENVIRONMENTS) {
         if (envReleases[env]) {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
Zee2413: Is this needed? Warning went away for me when I ran npm ci to get semver

satyakigh (dismissed):

src/services/CfnService.ts
```diff
@@ -60,12 +60,16 @@ import {
...
+import { SettingsManager } from '../settings/SettingsManager';
 import { DeploymentMode } from '../stacks/actions/StackActionRequestType';
 import { Count } from '../telemetry/TelemetryDecorator';
 import { AwsClient } from './AwsClient';
 
 export class CfnService {
-    public constructor(private readonly awsClient: AwsClient) {}
+    public constructor(
+        private readonly awsClient: AwsClient,
+        private readonly settingsManager: SettingsManager,
```
Zee2413: Can we directly use the default settings instead of injecting the settings manager? Similar to hover router: https://github.com/aws-cloudformation/cloudformation-languageserver/blob/main/src/hover/HoverRouter.ts#L30

PR #318, opened 2025-11-26
Reset extension between tests

satyakigh:
* Instantiating TestExtension is the slowest part of e2e tests
* Creating it once per test file and resetting the state speeds up tests
* Reduces tests duration by ~1 minute


tst/e2e/Completion.yaml.test.ts
```diff
@@ -7,11 +7,15 @@ describe('Completion Tests', () => {
     const documentUri = 'file:///test.yaml';
     let extension: TestExtension;
 
-    beforeEach(() => {
+    beforeAll(() => {
```
Zee2413: We can just move this above and set to `const extension`

Zee2413 (approved):

PR #357, opened 2025-12-09
Replace in-process mutex with cross-process file locking for FileStore

satyakigh:
### Summary for Reviewers

This PR refactors the EncryptedFileStore to improve reliability and fix potential data loss issues during concurrent access across multiple processes.

### Key Changes

1. File locking mechanism replaced: Switched from async-mutex (in-process only) to proper-lockfile (cross-process file locking) to properly handle concurrent access from multiple language server instances.

2. Eager file creation on startup: The store now creates encrypted files immediately on construction rather than lazily on first write. This ensures the lock file target exists before any operations.

3. Corrupted file recovery: Added automatic recovery when decryption fails - the store recreates itself with empty content instead of crashing.

4. Store initialization moved to factory constructor: FileStoreFactory now pre-creates all stores at startup with a configurable list of store names, and get() throws if requesting an unknown store (fail-fast behavior).

5. Simplified content management: Removed nullable content field - it's now always initialized. Each write operation re-reads the file under lock to ensure it has the latest data from other processes.

6. Replaced error stack parsing with simply adding the full stack (since these are not metric dimensions anymore)

### DIFF
--- a/src/datastore/file/EncryptedFileStore.ts
+++ b/src/datastore/file/EncryptedFileStore.ts
@@ -2,14 +2,15 @@ import { existsSync, readFileSync, statSync, writeFileSync } from 'fs'; // eslin
 import { writeFile } from 'fs/promises';
 import { join } from 'path';
 import { Logger } from 'pino';
-import { lock, lockSync } from 'proper-lockfile';
+import { lock, LockOptions, lockSync } from 'proper-lockfile';
 import { LoggerFactory } from '../../telemetry/LoggerFactory';
 import { ScopedTelemetry } from '../../telemetry/ScopedTelemetry';
 import { TelemetryService } from '../../telemetry/TelemetryService';
 import { DataStore } from '../DataStore';
 import { decrypt, encrypt } from './Encryption';
 
-const LOCK_OPTIONS = { stale: 10_000 }; // 10 seconds
+const LOCK_OPTIONS_SYNC: LockOptions = { stale: 10_000 };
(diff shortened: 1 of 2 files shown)


src/datastore/file/EncryptedFileStore.ts
```diff
@@ -1,122 +1,116 @@
...
         return {
-            entries: Object.keys(this.content ?? {}).length,
+            entries: Object.keys(this.content).length,
             totalSize: existsSync(this.file) ? statSync(this.file).size : 0,
         };
     }
+
+    private async withLock<T>(operation: string, fn: () => Promise<T>): Promise<T> {
+        return await this.telemetry.measureAsync(operation, async () => {
+            const release = await lock(this.file, LOCK_OPTIONS);
```
kddejong: Should this be in the try?  what if it doesn't get the lock do we just bubble the failure to the parent function so it can decide how to handle it?

satyakigh: ya then the write would fail and the caller would have to handle it, for the existing usage of persisted stores, it would retry to download schemas and save again

src/utils/Errors.ts
```diff
@@ -40,7 +38,7 @@ export function extractLocationFromStack(stack?: string): Record<string, string>
                 }
             }
 
-            return newLine;
+            return newLine.replaceAll('\\\\', '/').replaceAll('\\', '/');
```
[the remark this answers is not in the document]
satyakigh: Its an edge case for windows where some paths could have double `\\`

PR #364, opened 2025-12-15
Improve intrinsic functions and add json fallback parsing

satyakigh:
### Description

This PR improves how the CloudFormation Language Server parses and handles intrinsic functions (like !If, !Sub, !Ref, Fn::GetAtt, etc.) in both YAML and JSON
templates. It also adds a JSON fallback parser for malformed/incomplete documents.

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

#### 1. SyntaxTree.ts - Core parsing improvements

Intrinsic function path handling:
- Refactored YAML tag handling (!If, !Sub, !Ref, etc.) to run independently rather than in an else-if chain
- This ensures both the intrinsic function name AND the parent property are captured in the path
- Added support for flow_node in addition to block_node for YAML intrinsics
- Added check to skip when parent is ERROR node (lets fallback handle malformed trees)

New JSON fallback parser:
- Added pathAndEntityJsonFallback() method (~100 lines) for malformed JSON documents
- Uses text-based stack parsing to infer context when tree-sitter produces ERROR nodes
- Tracks JSON structure ({, }, [, ], :, ,) to build path context
- Enables completions in incomplete JSON templates (e.g., typing a property name)

Minor fix:
- Added handling for cursor after YAML TAG nodes (e.g., !Sub or incomplete !E)

#### 2. LogicalIdReferenceFinder.ts - Simplified reference extraction

- Simplified early-exit checks using text.includes() instead of indexOf()
- Removed separate JsonSub/YamlSub regex patterns - now uses unified SubVariables pattern
- Added !If and Fn::If: pattern matching for YAML
- Unified ${...} variable extraction for both JSON and YAML into single SubVariables regex
- Improved SubVariables regex to capture only the resource name from ${Resource.Attribute} syntax
- Added 'AWS' to CommonProperties exclusion set (filters out AWS::Region, etc.)

### DIFF
--- a/src/context/semantic/LogicalIdReferenceFinder.ts
+++ b/src/context/semantic/LogicalIdReferenceFinder.ts
@@ -98,6 +98,9 @@ function findYamlIntrinsicReferences(text: string, logicalIds: Set<string>): voi
     if (text.includes('Condition')) {
         extractMatches(text, YamlCondition, logicalIds);
     }
+    if (text.includes('!ValueOf')) {
+        extractMatches(text, YamlValueOfShort, logicalIds);
+    }
     if (text.includes('Fn::ValueOf')) {
         extractMatches(text, YamlValueOf, logicalIds);
     }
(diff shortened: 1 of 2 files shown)


src/context/semantic/LogicalIdReferenceFinder.ts
```diff
@@ -35,44 +36,33 @@ export function referencedLogicalIds(
...
-    const dependsIndex = text.indexOf('"DependsOn"');
-    const subVarIndex = text.indexOf('${');
-
-    if (refIndex !== -1) {
+    // Early exit checks - only run regex if marker exists
+    if (text.includes('"Ref"')) {
         extractMatches(text, JsonRef, logicalIds);
     }
-    if (getAttIndex !== -1) {
+    if (text.includes('"Fn::GetAtt"')) {
```
Zee2413: Would using `IntrinsicFunction` enum make these more maintainable?

satyakigh: A lof of the intrinsic functions are being used here, it was easier to read and follow without using the enums here

src/context/syntaxtree/SyntaxTree.ts
```diff
@@ -469,6 +478,48 @@ export abstract class SyntaxTree {
...
+            const parentIsError = NodeType.isNodeType(parent, CommonNodeTypes.ERROR);
+            if (
+                !parentIsError &&
+                this.type === DocumentType.YAML &&
+                (NodeType.isNodeType(current, YamlNodeTypes.BLOCK_NODE) ||
+                    NodeType.isNodeType(current, YamlNodeTypes.FLOW_NODE)) &&
+                current.children?.length > 0 &&
+                current.children.some((child) => NodeType.isNodeType(child, YamlNodeTypes.TAG))
+            ) {
+                const tagNode = current.children.find((child) => NodeType.isNodeType(child, YamlNodeTypes.TAG));
```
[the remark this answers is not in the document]
satyakigh: Both of these exit early since it just looks for the first child and if we compute this before the if, then the search always runs even if the other conditions might be false. This is only looking for first element so it should be fast enough

Zee2413 (approved):

PR #369, opened 2025-12-18
Fix intrinsic parsing issues

satyakigh:
### Bug 1: Nested !Ref Inside Intrinsics Returns Wrong Completions

Scenario: User types !Equals [!Ref AWS::, ...]

Problem: The completion provider saw the outer !Equals function and didn't recognize the nested !Ref. It incorrectly suggested intrinsic function names instead
of pseudo-parameters.

Fix: Detect when context.text starts with !Ref and route to the Ref argument handler. Also fixed shouldUseIntrinsicFunctionProvider() to only trigger when the
user is actually typing a function name (no space in text, or text after last space starts with !).

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

### Bug 2: !Condition Completions Not Working

Scenario: User types !Condition IsProd to reference a condition

Problem: Tree-sitter places the cursor inside the argument text, not at the !Condition tag. The existing check only looked at property paths and entitySection,
so it failed to recognize the condition context.

Fix: Added isInsideConditionShortForm() that traverses up to 3 parent nodes to find if we're inside a !Condition expression. Also strip the !Condition prefix
from search text before fuzzy matching.

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

### Bug 3: Go-to-Definition Broken in !Sub Strings

Scenario: User clicks on ${EnvironmentName} inside a !Sub string to navigate to the parameter definition

Problem: context.text was empty or contained the full ${...} pattern. The definition provider couldn't extract the variable name.

Fix: Added fallback logic to extract the variable name from entityRootNode.text using regex when context.text is empty or contains ${.

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

### Bug 4: Regex False Positives in Reference Finder

Problem 1: YamlListItem pattern -\s*([A-Za-z]...) matched hyphens in resource names like my-bucket.

Fix: Require whitespace before the hyphen: /(?:^|\s)-\s+([A-Za-z][A-Za-z0-9]*)/gm

Problem 2: SubVariables pattern required ., :, or } immediately after the variable name, missing valid cases.

Fix: Match any non-alphanumeric terminator: /\$\{([A-Za-z][A-Za-z0-9]*)(?:[^A-Za-z0-9]|$)/g

### DIFF
--- a/src/autocomplete/CompletionRouter.ts
+++ b/src/autocomplete/CompletionRouter.ts
@@ -1,8 +1,11 @@
+import { SyntaxNode } from 'tree-sitter';
 import { CompletionItem, CompletionParams } from 'vscode-languageserver';
 import { Context } from '../context/Context';
 import { ContextManager } from '../context/ContextManager';
 import {
+    EntitySection,
     IntrinsicFunction,
+    IntrinsicShortForms,
     IntrinsicsUsingConditionKeyword,
     ResourceAttribute,
     TopLevelSection,
@@ -38,7 +41,6 @@ export type CompletionProviderType =
     | 'IntrinsicFunctionArgument'
     | 'ParameterTypeValue'
     | EntityType;
-const Condition = 'Condition';
 
 export class CompletionRouter implements SettingsConfigurable, Closeable {
     private completionSettings: CompletionSettings = DefaultSettings.completion;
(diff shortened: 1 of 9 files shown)


src/autocomplete/CompletionRouter.ts
```diff
@@ -137,31 +160,43 @@ export class CompletionRouter implements SettingsConfigurable, Closeable {
...
+                ResourceAttribute.UpdatePolicy,
+                EntitySection.Condition,
+            )
+        ) {
             return true;
         }
 
-        // Resource Metadata Condition: ['Resources', 'LogicalId', 'Metadata', this.CONDITION]
-        if (context.matchPathWithLogicalId(TopLevelSection.Resources, ResourceAttribute.Metadata, Condition)) {
+        // Resource Metadata Condition: ['Resources', 'LogicalId', 'Metadata', 'Condition']
```
kddejong: This change is good but I don't think a Resource Metadata Condition is a thing.

PR #373, opened 2025-12-23
Fix LMDB errors and add recovery mechanisms

satyakigh:
### 1. MDB_BAD_RSLOT (Fork Detection Issues)

**Error Message:**
```
MDB_BAD_RSLOT: Invalid reuse of reader locktable slot: 
The reader lock pid X, txn Y, doesn't match env pid Z
```

**Cause:**
LMDB uses memory-mapped files with process-specific reader lock slots. When a process forks (common in VS Code's extension host), the child process inherits the parent's LMDB environment but the reader lock table still references the parent's PID. Any read operation in the child triggers this error.

---

### 2. MDB_CORRUPTED / MDB_PAGE_NOTFOUND

**Error Messages:**
```
MDB_CORRUPTED: Located page was wrong type
MDB_PAGE_NOTFOUND: Requested page not found
```

**Cause:**
- Stale memory mappings after fork
- Incomplete writes from crashes
- Concurrent access issues across processes
- File system corruption

These errors indicate the B+ tree structure is inconsistent - either a page pointer references the wrong page type, or a referenced page doesn't exist.
---

### 3. Closed Database Errors

**Error Message:**
```
Can not renew a transaction from a closed database
```

**Cause:**
Operations attempted on an LMDB environment that was closed (either explicitly or due to error recovery). The `renewReadTxn` function tries to reuse a transaction from an invalidated environment.

---

### 4. MDB_BAD_TXN

**Error Message:**
```
MDB_BAD_TXN: Transaction must abort, has a child, or is invalid
```

**Cause:**
Transaction state corruption, often from:
- Using transactions across fork boundaries
- Nested transaction issues
- Cursor operations on aborted transactions

### Fix 1: Proactive Fork Detection
- Tracks the PID when environment was opened (`openPid`)
- Before every database operation, checks if current PID differs
- If fork detected, proactively reopens environment before the operation
- Updates all store handles with fresh database references

**Addresses:** MDB_BAD_RSLOT errors (2,856 occurrences)

---

### Fix 2: Reactive Error Recovery
- Catches errors from LMDB operations
- Classifies errors into fork-related vs corruption-related
- Triggers appropriate recovery strategy

---

### Fix 3: Environment Recreation
- Creates fresh LMDB environment with new memory mappings
- Updates PID tracking
- Recreates all database handles
- Existing store references remain valid (they get updated handles)

---

### Fix 4: Store Handle Updates
- Allows in-place update of the underlying database handle
- Callers holding store references don't need new references
- Enables seamless recovery without breaking existing code

---

### Fix 5: Operation Wrapping
- Every operation calls `validateDatabase()` first (fork detection)
- If operation fails, `onError()` triggers recovery
- Telemetry wrapping provides observability

### DIFF
--- a/src/schema/GetSchemaTask.ts
+++ b/src/schema/GetSchemaTask.ts
@@ -5,6 +5,7 @@ import { CfnService } from '../services/CfnService';
 import { LoggerFactory } from '../telemetry/LoggerFactory';
 import { ScopedTelemetry } from '../telemetry/ScopedTelemetry';
 import { Measure, Telemetry } from '../telemetry/TelemetryDecorator';
+import { classifyAwsError } from '../utils/AwsErrorMapper';
 import { AwsRegion } from '../utils/Region';
 import { downloadFile } from '../utils/RemoteDownload';
 import { PrivateSchemas, PrivateSchemasType, PrivateStoreKey } from './PrivateSchemas';
@@ -73,6 +74,9 @@ export class GetPublicSchemaTask extends GetSchemaTask {
 export class GetPrivateSchemasTask extends GetSchemaTask {
     private readonly logger = LoggerFactory.getLogger(GetPrivateSchemasTask);
 
+    @Telemetry()
+    private readonly telemetry!: ScopedTelemetry;
+
     constructor(private readonly getSchemas: () => Promise<DescribeTypeOutput[]>) {
         super();
     }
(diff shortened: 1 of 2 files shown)


atennak1 (approved):

src/datastore/lmdb/LMDBStore.ts
```diff
@@ -1,53 +1,72 @@
...
 
-    get<T>(key: string): T | undefined {
-        return this.telemetry.countExecution('get', () => {
-            return this.store.get(key) as T | undefined;
+    updateStore(store: Database<unknown, string>) {
+        this.store = store;
+    }
+
+    private exec<T>(op: string, fn: () => T): T {
+        return this.telemetry.measure(op, () => {
```
Zee2413: All the prior `countExecution` are now `measure`. Are the added metrics needed?

Zee2413 (approved):

PR #375, opened 2025-12-26
Refresh indentation after user types in an empty template

deepfuriya:
## Description

Fixes indentation detection for templates that start empty. Previously, when a user opened a new empty
file and started typing with their preferred indentation (e.g., 2 spaces), the language server would use
the fallback tab size instead of detecting the user's actual indentation style.

## Changes

- Added lazy indentation detection that re-evaluates when getTabSize() is called, allowing detection
after content is added to an initially empty document
- Track whether indentation has been successfully detected to avoid redundant re-detection
- Pass detectIndentation setting through to getTabSize() to respect user preferences

## Testing

- Added unit tests for Document indentation detection (lazy detection, fallback behavior, detection
caching)
- Added unit tests for DocumentManager.getEditorSettingsForDocument() to validate VS Code settings
integration
- Added E2E tests verifying snippet indentation matches user-typed indentation in completions

### DIFF
--- a/tst/e2e/Completion.yaml.test.ts
+++ b/tst/e2e/Completion.yaml.test.ts
@@ -1,5 +1,5 @@
 import { beforeEach, describe, expect, test, afterAll } from 'vitest';
-import { CompletionList } from 'vscode-languageserver';
+import { CompletionList, InsertTextFormat } from 'vscode-languageserver';
 import { TestExtension } from '../utils/TestExtension';
 import { WaitFor } from '../utils/Utils';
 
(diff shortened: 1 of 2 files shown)


src/document/Document.ts
```diff
@@ -160,10 +161,25 @@ export class Document {
...
+        }
         return this.tabSize;
     }
 
+    private refreshIndentationIfNeeded(): void {
+        if (this.indentationDetected) {
+            return;
+        }
+
+        const detected = this.detectIndentationFromContent();
```
satyakigh: Whats the difference between `processIndentation` and `detectIndentationFromContent`

deepfuriya: `processIndentation` is a public method that will update the `tabSize` based on the either detecting the indentation by scanning the template or use the fallback value if detect indentation setting is disabled. 

`detectIndentationFromContent` is the private method used for scanning the first 30 lines of file.

chrisqm-dev (approved):

PR #376, opened 2025-12-29
Exclude Fn::ForEach resources from Ref and GetAtt argument completions

deepfuriya:
## Summary
Filters out Fn::ForEach::* resources from autocomplete suggestions when inside !Ref and !GetAtt intrinsic functions. These loop construct resources are not valid references and should not appear as completion options.

## Changes
- IntrinsicFunctionArgumentCompletionProvider.ts:
  - getResourceCompletions(): Skip resources with names starting with Fn::ForEach::
  - getGetAttResourceCompletions(): Skip resources with names starting with Fn::ForEach::

## Testing
- Added unit tests in IntrinsicFunctionArgumentCompletionProvider.Ref.test.ts
- Added unit tests in IntrinsicFunctionArgumentCompletionProvider.GetAtt.test.ts
- Added e2e test in Completion.test.ts for YAML cross-reference completions

### DIFF
--- a/src/autocomplete/IntrinsicFunctionArgumentCompletionProvider.ts
+++ b/src/autocomplete/IntrinsicFunctionArgumentCompletionProvider.ts
@@ -341,6 +341,11 @@ export class IntrinsicFunctionArgumentCompletionProvider implements CompletionPr
                 continue;
             }
 
+            // Skip Fn::ForEach resources
+            if (resourceName.startsWith(IntrinsicFunction.ForEach)) {
+                continue;
+            }
+
             const resource = resourceContext.entity;
 
             completionItems.push(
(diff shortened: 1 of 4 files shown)


src/autocomplete/IntrinsicFunctionArgumentCompletionProvider.ts
```diff
@@ -341,6 +341,11 @@ export class IntrinsicFunctionArgumentCompletionProvider implements CompletionPr
                 continue;
             }
 
+            // Skip Fn::ForEach resources
+            if (resourceName.startsWith('Fn::ForEach::')) {
```
satyakigh: can the constants from ContextType be used?

deepfuriya: yes, updated it.

PR #387, opened 2026-01-07
Fix goto assertions

satyakigh:
* Goto position expectations were not being asserted

### DIFF
--- a/tst/utils/TemplateBuilder.ts
+++ b/tst/utils/TemplateBuilder.ts
@@ -432,19 +432,7 @@ export class TemplateBuilder {
 
         const actualLabels = completions.items.map((item) => item.label);
         if (expected.items) {
-            // Check if all expected items are present
-            for (const expectedItem of expected.items) {
-                if (!actualLabels.includes(expectedItem)) {
-                    throw new Error(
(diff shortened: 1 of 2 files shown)


tst/utils/TemplateBuilder.ts
```diff
@@ -691,6 +679,13 @@ export class TemplateBuilder {
                     );
                 }
             }
+
+            if (expected.targetPosition !== undefined) {
```
satyakigh: Target positions were not being checked

chrisqm-dev (approved):

tst/utils/TemplateBuilder.ts
```diff
@@ -691,6 +679,13 @@ export class TemplateBuilder {
...
+                const firstDef = definitionArray[0];
+                expectAt(firstDef.range.start, position, `Target position mismatch${desc}`).toEqual(
```
Zee2413: Since we are fixing this, consider adding in end position as well to assert the correct range is provided

satyakigh: We don't have the end positions defined, the resources are over multiple rows/lines

PR #445, opened 2026-02-18
fix: handle corrupted LMDB database during runtime

kddejong:
## Fix: Delete corrupted LMDB database during recovery

Fixes aws/aws-toolkit-vscode#8350

### Problem

Users encounter `MDB_CORRUPTED: Located page was wrong type` errors that prevent the CloudFormation language server from initializing. The error occurs during the LSP initialize request when the language server tries to access a corrupted LMDB database.

**Error message:**

Server initialization failed.
Message: Request initialize failed with message: MDB_CORRUPTED: Located page was wrong type

The existing error handler attempted recovery by reopening the database, but it didn't delete the corrupted files. This meant:
1. Language server starts and opens the corrupted database (succeeds - LMDB doesn't validate on open)
2. LSP initialize request triggers an operation that reads from the database
3. `MDB_CORRUPTED` error is thrown
4. Recovery handler reopens the **same corrupted database**
5. Next operation fails with the same error
6. **Language server becomes unusable** - users cannot use CloudFormation features

### Solution

Modified `recoverFromCorruption()` to delete the corrupted database directory before reopening. This allows LMDB to create a fresh, empty database and recover from corruption.

**Changes:**
- Added `isCorruptionError()` function to identify corruption errors (`MDB_CORRUPTED`, `MDB_PAGE_NOTFOUND`, `MDB_PANIC`)
- Modified `recoverFromCorruption()` to delete the database directory before calling `reopenEnv()`
- Added tests for corruption error detection and recovery behavior

### Trade-offs

**Data Loss:** When corruption is detected, the database is deleted and recreated empty. This means:
- Cached schemas are lost and will be re-downloaded
- Any other cached data is lost

This is acceptable because:
- The alternative is a completely non-functional language server
- The data is cache that can be regenerated
- Users prefer a working language server over preserving corrupted cache

**Scope:** This fix handles runtime corruption (errors during operations). Severe corruption that causes LMDB to crash at the C level cannot be caught, but the user's reported error is a catchable JavaScript exception.

### Testing

Added 6 new tests covering:
- Corruption error message detection (positive cases)
- Non-corruption error detection (negative cases)  
- Database deletion behavior

**Note:** We cannot test actual corruption recovery in automated tests because:
- Random corruption either works fine or causes uncatchable process crashes
- `MDB_CORRUPTED` errors occur during specific internal LMDB operations that are difficult to trigger reliably
- Attempting to reopen LMDB in the same process during tests causes worker hangs

However, the fix is sound because:
1. The user's error is a catchable JavaScript exception (confirmed from issue)
2. The old code had the error handler but didn't delete files (confirmed by code review)
3. Our fix adds the missing deletion step before reopening

### Verification

All existing tests pass (40 total). The fix addresses the root cause: not deleting corrupted database files before attempting recovery.

### DIFF
--- a/src/datastore/LMDB.ts
+++ b/src/datastore/LMDB.ts
@@ -97,17 +97,8 @@ export class LMDBStoreFactory implements DataStoreFactory {
 
         if (msg.includes('MDB_BAD_RSLOT') || msg.includes("doesn't match env pid")) {
             this.recoverFromFork();
-        } else if (isCorruptionError(error)) {
-            this.recoverFromCorruption();
-        } else if (
-            msg.includes('MDB_CURSOR_FULL') ||
-            msg.includes('MDB_BAD_TXN') ||
-            msg.includes('Commit failed') ||
-            msg.includes('closed database')
-        ) {
-            this.telemetry.count('transient', 1);
-            this.log.warn('Transient error detected, reopening LMDB');
-            this.reopenEnv();
+        } else {
+            this.recoverFromError();
         }
     }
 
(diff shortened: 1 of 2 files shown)


Zee2413 (approved):
Would be good to add test case for transient error handling.

PR #451, opened 2026-02-24
Add error message to resource importer result

gemammercado:
This PR adds an error message to the resource importer result to provide more context about why the import failed.
Client side change: https://github.com/aws/aws-toolkit-vscode/pull/8632

### DIFF
--- a/src/resourceState/ResourceStateImporter.ts
+++ b/src/resourceState/ResourceStateImporter.ts
@@ -16,6 +16,7 @@ import { CfnLspProviders } from '../server/CfnLspProviders';
 import { LoggerFactory } from '../telemetry/LoggerFactory';
 import { ScopedTelemetry } from '../telemetry/ScopedTelemetry';
 import { Telemetry, Measure } from '../telemetry/TelemetryDecorator';
+import { extractErrorMessage } from '../utils/Errors';
 import { getIndentationString } from '../utils/IndentationUtils';
 import { ResourceStateManager } from './ResourceStateManager';
 import {
@@ -160,7 +161,6 @@ export class ResourceStateImporter {
             completionItem: undefined,
             failedImports: {},
             successfulImports: {},
-            failureReasons: {},
         };
 
         const generatedLogicalIds = new Set<string>();
@@ -198,7 +198,7 @@ export class ResourceStateImporter {
                 } catch (error) {
                     log.error(error, `Error importing resource state for ${resourceType} id: ${resourceIdentifier}`);
                     this.getOrCreate(importResult.failedImports, resourceType, []).push(resourceIdentifier);
-                    const errorMessage = error instanceof Error ? error.message : String(error);
+                    const errorMessage = extractErrorMessage(error);
                     importResult.failureReasons ??= {};
                     importResult.failureReasons[resourceType] ??= {};
                     importResult.failureReasons[resourceType][resourceIdentifier] = errorMessage;
--- a/src/resourceState/ResourceStateImporter.ts
+++ b/src/resourceState/ResourceStateImporter.ts
@@ -177,6 +177,10 @@ export class ResourceStateImporter {
             for (const resourceIdentifier of resourceSelection.resourceIdentifiers) {
                 try {
                     const resourceState = await this.resourceStateManager.getResource(resourceType, resourceIdentifier);
+                    if (!resourceState) {
+                        this.getOrCreate(importResult.failedImports, resourceType, []).push(resourceIdentifier);
+                        continue;
+                    }
                     this.getOrCreate(importResult.successfulImports, resourceType, []).push(resourceIdentifier);
                     const logicalId = this.generateUniqueLogicalId(
                         resourceType,
--- a/src/resourceState/ResourceStateImporter.ts
+++ b/src/resourceState/ResourceStateImporter.ts
@@ -160,6 +160,7 @@ export class ResourceStateImporter {
             completionItem: undefined,
             failedImports: {},
             successfulImports: {},
+            failureReasons: {},
         };
 
         const generatedLogicalIds = new Set<string>();
@@ -175,8 +176,8 @@ export class ResourceStateImporter {
             }
             for (const resourceIdentifier of resourceSelection.resourceIdentifiers) {
                 try {
-                    const resourceState = await this.resourceStateManager.getResource(resourceType, resourceIdentifier);
-                    if (resourceState) {
+                    const result = await this.resourceStateManager.getResource(resourceType, resourceIdentifier);
+                    if (result.resource) {
                         this.getOrCreate(importResult.successfulImports, resourceType, []).push(resourceIdentifier);
                         const logicalId = this.generateUniqueLogicalId(
                             resourceType,
@@ -192,7 +193,7 @@ export class ResourceStateImporter {
                                 DeletionPolicy:
                                     purpose === ResourceStatePurpose.IMPORT ? DeletionPolicyOnImport : undefined,
                                 Properties: this.applyTransformations(
-                                    resourceState.properties,
+                                    result.resource.properties,
                                     schema,
                                     purpose,
                                     logicalId,
@@ -202,10 +203,19 @@ export class ResourceStateImporter {
                         });
                     } else {
                         this.getOrCreate(importResult.failedImports, resourceType, []).push(resourceIdentifier);
+                        if (result.error) {
+                            importResult.failureReasons ??= {};
+                            importResult.failureReasons[resourceType] ??= {};
+                            importResult.failureReasons[resourceType][resourceIdentifier] = result.error;
+                        }
                     }
                 } catch (error) {
                     log.error(error, `Error importing resource state for ${resourceType} id: ${resourceIdentifier}`);
                     this.getOrCreate(importResult.failedImports, resourceType, []).push(resourceIdentifier);
+                    const errorMessage = error instanceof Error ? error.message : String(error);
+                    importResult.failureReasons ??= {};
+                    importResult.failureReasons[resourceType] ??= {};
+                    importResult.failureReasons[resourceType][resourceIdentifier] = errorMessage;
                 }
             }
         }
--- a/src/resourceState/ResourceStateImporter.ts
+++ b/src/resourceState/ResourceStateImporter.ts
@@ -157,6 +157,7 @@ export class ResourceStateImporter {
         parentResourceType?: string,
     ): Promise<{ fetchedResourceStates: ResourceTemplateFormat[]; importResult: ResourceStateResult }> {
         const fetchedResourceStates: ResourceTemplateFormat[] = [];
+        const failureReasons: Record<string, Record<string, string>> = {};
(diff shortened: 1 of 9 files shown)


satyakigh (dismissed):

src/resourceState/ResourceStateImporter.ts
```diff
@@ -176,36 +177,31 @@ export class ResourceStateImporter {
...
+                            DeletionPolicy:
+                                purpose === ResourceStatePurpose.IMPORT ? DeletionPolicyOnImport : undefined,
+                            Properties: this.applyTransformations(resourceState.properties, schema, purpose, logicalId),
+                            Metadata: await this.createMetadata(resourceIdentifier, purpose),
+                        },
+                    });
                 } catch (error) {
                     log.error(error, `Error importing resource state for ${resourceType} id: ${resourceIdentifier}`);
                     this.getOrCreate(importResult.failedImports, resourceType, []).push(resourceIdentifier);
+                    const errorMessage = error instanceof Error ? error.message : String(error);
```
satyakigh: use `extractErrorMessage`

src/resourceState/ResourceStateManager.ts
```diff
@@ -75,20 +76,20 @@ export class ResourceStateManager implements SettingsConfigurable, Closeable {
         } catch (error) {
             if (error instanceof ResourceNotFoundException) {
                 log.info(`No resource found for type ${typeName} and identifier "${identifier}"`);
-                return;
+            } else {
+                log.error(error, `CCAPI GetResource failed for type ${typeName} and identifier "${identifier}"`);
             }
+
             if (isClientError(error)) {
```
satyakigh: Is this line doing anything anymore? Since there is no more return - it just logs

src/resourceState/ResourceStateImporter.ts
```diff
@@ -160,6 +160,7 @@ export class ResourceStateImporter {
             completionItem: undefined,
             failedImports: {},
             successfulImports: {},
+            failureReasons: {},
```
satyakigh: Does this have to be initialized? the type says its optional

tst/unit/autocomplete/ResourceStateCompletionProvider.test.ts
```diff
@@ -115,7 +115,7 @@ describe('ResourceStateCompletionProvider', () => {
         });
 
         mockComponents.schemaRetriever.getDefault.returns(s3Schemas);
-        mockComponents.resourceStateManager.getResource.resolves(undefined);
+        mockComponents.resourceStateManager.getResource.resolves();
```
satyakigh: Is this being mocked correctly? Isn't the return type of `getResource` an actual object?

satyakigh (commented):
No tests for the new failure reason attributes

src/resourceState/ResourceStateManager.ts
gemammercado: No this error will surface on client side anyway so I'm removing

tst/unit/autocomplete/ResourceStateCompletionProvider.test.ts
gemammercado: Changed the `.resolves()` / `.resolves(undefined)` to `.rejects(new Error('Resource not found'))` because `getResource` now throws when a resource isn't found instead of returning undefined.

src/resourceState/ResourceStateImporter.ts
gemammercado: Removed. `failureReasons` is now lazily initialized with `??=` only when an error occurs, so it stays `undefined` when all imports succeed

src/resourceState/ResourceStateManager.ts
```diff
@@ -59,13 +58,14 @@ export class ResourceStateManager implements SettingsConfigurable, Closeable {
         this.initializeCounters();
     }
 
-    @Measure({ name: 'getResource', captureErrorType: true })
-    public async getResource(typeName: ResourceType, identifier: ResourceId): Promise<ResourceState | undefined> {
+    @Measure({ name: 'getResource' })
```
Zee2413: why are we getting rid of `captureErrorType`?

src/resourceState/ResourceStateManager.ts
```diff
@@ -75,20 +75,16 @@ export class ResourceStateManager implements SettingsConfigurable, Closeable {
...
-                return;
+            } else {
+                log.error(error, `CCAPI GetResource failed for type ${typeName} and identifier "${identifier}"`);
             }
             throw error;
         }
 
         if (!output?.TypeName || !output?.ResourceDescription?.Identifier || !output?.ResourceDescription?.Properties) {
-            log.error(
+            throw new Error(
```
Zee2413: these 400 level errors would get reported as faults

src/resourceState/ResourceStateManager.ts
gemammercado: mistake from when I was resolving merge conflicts, adding back

src/resourceState/ResourceStateManager.ts
gemammercado: reverting to old format where we log and return ;

src/resourceState/ResourceStateManager.ts
```diff
@@ -75,11 +75,8 @@ export class ResourceStateManager implements SettingsConfigurable, Closeable {
         } catch (error) {
             if (error instanceof ResourceNotFoundException) {
                 log.info(`No resource found for type ${typeName} and identifier "${identifier}"`);
-                return;
-            }
-            if (isClientError(error)) {
-                log.info(`Client error for type ${typeName} and identifier "${identifier}"`);
-                return;
+            } else {
```
Zee2413: Why this change? Why are we throwing everything now?

src/resourceState/ResourceStateImporter.ts
```diff
@@ -176,36 +177,35 @@ export class ResourceStateImporter {
...
+                                purpose === ResourceStatePurpose.IMPORT ? DeletionPolicyOnImport : undefined,
+                            Properties: this.applyTransformations(resourceState.properties, schema, purpose, logicalId),
+                            Metadata: await this.createMetadata(resourceIdentifier, purpose),
+                        },
+                    });
                 } catch (error) {
                     log.error(error, `Error importing resource state for ${resourceType} id: ${resourceIdentifier}`);
                     this.getOrCreate(importResult.failedImports, resourceType, []).push(resourceIdentifier);
+                    const errorMessage = extractErrorMessage(error);
+                    importResult.failureReasons ??= {};
```
Zee2413: `importResult` can be initialized with `failureReasons` as `{}` in line 160 and this check will not be needed.

gemammercado: Typescript throws an error either way: `'importResult.failureReasons' is possibly 'undefined'.`

src/resourceState/ResourceStateManager.ts
Zee2413: The `ResourceNotFoundException` and `isClientError` special handling was put in place to omit faults from the getResource metric. Throwing from here is just going to undo all that work and we will see the faults spike up.

gemammercado: These are errors that used to return undefined and thus showed no error message to the client

gemammercado: Ah i see

Zee2413: I understand the intent to surface the errors to clients. You'll have to design a solution that achieves that and does not regress on the work to scope down the fault metric.

gemammercado: I will just make the return value an object and instead of throwing on these cases, we will return the error message

src/resourceState/ResourceStateManager.ts
```diff
@@ -75,11 +75,8 @@ export class ResourceStateManager implements SettingsConfigurable, Closeable {
         } catch (error) {
             if (error instanceof ResourceNotFoundException) {
                 log.info(`No resource found for type ${typeName} and identifier "${identifier}"`);
```
gemammercado, line 81: we can return nothing here because ResourceNotFound message is rendered [client side](https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/awsService/cloudformation/ui/resourceSelector.ts#L117)

Zee2413: That is not related to the `ResourceNotFoundException` here. That's just showing a client message when nothing is found which could be caused by *anything* on the server side. This is specifically a not found exception from CCAPI

src/resourceState/ResourceStateManager.ts
```diff
@@ -60,12 +63,13 @@ export class ResourceStateManager implements SettingsConfigurable, Closeable {
     }
 
     @Measure({ name: 'getResource', captureErrorType: true })
-    public async getResource(typeName: ResourceType, identifier: ResourceId): Promise<ResourceState | undefined> {
+    public async getResource(typeName: ResourceType, identifier: ResourceId): Promise<GetResourceResult> {
```
Zee2413, line 66: Are any of the caller behaviors changed that rely on `undefined`?

src/resourceState/ResourceStateManager.ts
```diff
@@ -75,20 +79,20 @@ export class ResourceStateManager implements SettingsConfigurable, Closeable {
         } catch (error) {
             if (error instanceof ResourceNotFoundException) {
                 log.info(`No resource found for type ${typeName} and identifier "${identifier}"`);
-                return;
-            }
-            if (isClientError(error)) {
+                return {};
```
Zee2413: return `{}` is defeating the purpose of the newly added `error` field in `GetResourceResult` type.

This branch occurs when a certain identifier actually does not exist in the aws account. That kind of information is exactly what would be meaningful to as a client message.

src/resourceState/ResourceStateManager.ts
gemammercado: Yep, all callers were updated. They are:
  
  1. `ResourceStateImporter.ts` before checked `if (resourceState)` -> Now checks `if(!result.resource).`
  
  2. ResourceStateCompletionProvider.ts -- before checked `if (!properties)` after `resourceState?.properties` -> `if (!result.resource)` and accesses `result.resource.properties`
  
  3. `ResourceStateManager.ts` (internal searchResourceByIdentifier) -- before caught all errors to determine "not found" ->`if (!result.resource)` for returned errors, and still catches thrown server errors.
  
No caller relies on undefined anymore, they all use the `{ resource?, error? }` shape.

src/resourceState/ResourceStateManager.ts
gemammercado: discussed offline, returning error message here and adding failureReasons to SearchResourceResult: https://github.com/aws/aws-toolkit-vscode/pull/8632/changes/741c00b5d953147ac3ed6f27e16d3fff375147cd

Zee2413 (approved):

satyakigh (approved):

Issue #450, opened 2026-02-24
[Bug]: Resource type completions are not displayed in JSON


Ben-Wormald:
### Describe the bug

Expected resource type completions are not displayed by editors (observed in VS Code and Zed) in JSON CloudFormation templates. This seems to be because the returned range for each completion (specifically `textEdit.range.end.character`) is longer than expected and is rejected by the editors.

### Steps to reproduce

1. Create a basic JSON CloudFormation template with an empty resource entry:

```json
{
  "AWSTemplateFormatVersion": "2010-09-09",
  "Resources": {
    "MyTestResource": {
    }
  }
}
```

2. Trigger a resource type completion:

```json
{
  "AWSTemplateFormatVersion": "2010-09-09",
  "Resources": {
    "MyTestResource": {
      "Type": "
    }
  }
}
```

3. No completions are suggested

### Expected behavior

Completions are suggested with available resources (`AWS::ACMPCA::Certificate` etc.)

### CloudFormation template (if applicable)

```yaml
{
  "AWSTemplateFormatVersion": "2010-09-09",
  "Resources": {
    "MyTestResource": {
      "Type": ""
    }
  }
}
```

### Language Server version

1.4.0

### Editor/IDE

VSCode 1.109.5, Zed 0.224.11

### Operating System

macOS

### Node.js version

v24.13.0

### Logs

```text
This is from Zed's logs, no relevant logs from the language server itself:


INFO  [project::lsp_command] completion out of expected range, start: PointUtf16 { row: 14, column: 8 }, end: PointUtf16 { row: 14, column: 8 }, range: Unclipped(PointUtf16 { row: 14, column: 8 })..Unclipped(PointUtf16 { row: 14, column: 9 })
```

kddejong:
This should be fixed in #450 it will be bundled in the next release.

PR #453, opened 2026-02-25
Fix resource type completion for JSON

Ben-Wormald:
*Issue #, if available:* https://github.com/aws-cloudformation/cloudformation-languageserver/issues/450

*Description of changes:*
- Fix replacement range being set too high
- Keep object keys when completing values e.g. resource types
- Simplify JSON formatting function for future improvements

There is still some behaviour for JSON completions that isn't ideal but hopefully this is a first steo to improving it, cheers


kddejong:
Closed in favor of #520

PR #511, opened 2026-03-31
Add LSP Client and long running tests in tools

chrisqm-dev:
*Description of changes:*
- Add lsp client methods to start up and interact with the cfn language server
 - Can be used for telemetry generator as well
- Add long-running tests
  - This tests the standalone.js and ensures it is usable for extended periods of time
  - Tests the following:
    - Completion on document updates
    - Hover on document updates
    - Region switching, and therefore public schema retrieval
    - Ensures performance doesn't degrade over time


tools/long-running/Templates.ts
```diff
@@ -0,0 +1,36 @@
+import { readFileSync } from 'fs';
+import { join } from 'path';
+
+const cache = new Map<string, string>();
+
+const loadTemplate = (name: string): string => {
```
satyakigh: The template utils from src can be used directly

tools/lsp-client/types.ts
```diff
@@ -0,0 +1,49 @@
...
+    clientId?: string;
+    storageDir?: string;
+    env?: NodeJS.ProcessEnv;
+    initTimeout?: number;
+    telemetryEnabled?: boolean;
+    featureFlags?: FeatureFlagType;
+    suppressLogLevels?: string[];
+}
+
+export interface ReadinessFlags {
```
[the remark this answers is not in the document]
chrisqm-dev: Modified to use all the src types

tools/lsp-client/LspClient.ts
```diff
@@ -0,0 +1,313 @@
...
+                    throw new Error('Lint and Guard services not ready yet');
+                }
+                console.log('Lint and Guard services are ready');
+            },
+            timeoutMs,
+            500, // Check every 500ms
+        );
+    }
+
+    get readiness(): ReadinessFlags {
```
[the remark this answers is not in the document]
chrisqm-dev: This should've been removed originally, removed now

tools/long-running/Monitoring.ts
```diff
@@ -0,0 +1,129 @@
+import { OperationType, getTesterConfig } from './testers/TesterTypes';
+
+export type StandaloneTestMetrics = {
+    operationsAttempted: number;
+    operationsFailed: number;
+    averageDuration: number | null;
```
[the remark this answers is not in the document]
chrisqm-dev: Because this test could run for a very long time and there are thousands of operations per minute, the duration array could get very large. We should keep the min, max, and average attributes to avoid having to calculate from the array. Added the durations array in case there's any other operations we use in the future, such as median or p99.

tools/long-running/run-long-running-test.ts
```diff
@@ -0,0 +1,20 @@
...
+    } catch (error) {
+        generateFinalReport(Date.now());
+        console.error('Standalone test failed:', error);
+        throw error;
+    } finally {
+        await orchestrator.cleanup();
+    }
+}
+
+void main();
```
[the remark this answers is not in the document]
chrisqm-dev: Added logic so that the exception is not swallowed in the event an exception (such as hover validation failure) occurs, and it fails fast. In the catch, we will explicitly throw the error

tools/lspClient/LspClient.ts
```diff
@@ -0,0 +1,313 @@
...
+
+        this.serverProcess = spawn('node', [this.config.serverPath, ...args], {
+            stdio: this.config.mode === 'ipc' ? ['pipe', 'pipe', 'pipe', 'ipc'] : ['pipe', 'pipe', 'pipe'],
+            env: { ...process.env, ...this.config.env },
+        });
+
+        console.log(`LspClient: Server process spawned with PID: ${this.serverProcess.pid}`);
+
+        // 2. Setup output monitoring for readiness detection
+        this.attachOutputListeners();
```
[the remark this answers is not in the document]
chrisqm-dev: Using lsp handler for system status

tools/stability/testers/CompletionTester.ts
```diff
@@ -0,0 +1,55 @@
...
+            if (!labels.has(required)) {
+                throw new Error(`${context} missing ${required}`);
+            }
+        }
+    }
+
+    async testAllScenarios(uri: string): Promise<void> {
+        // Test 1: Top-level completion using full document update
+        const basicTemplate = `AWSTemplateFormatVersion: '2010-09-09'
+`;
```
rongyamazon, line 25: nit: remove extra line

rongyamazon (dismissed):

tools/stability/testers/CompletionTester.ts
chrisqm-dev: the '`' is placed on a new line so that the completion provider returns all the top level resources, since the completion provider will only return these values on a new line

.gitignore
```diff
@@ -9,7 +9,8 @@
 *.node
 .DS_Store
 tools/*
-!tools/*.ts
```
satyakigh, line 12: Just need `!tools/**/*.ts`

tools/lspClient/LspClient.ts
```diff
@@ -0,0 +1,303 @@
...
+            console.error('LspClient: LSP handshake failed:', error);
+            throw error;
+        }
+    }
+
+    private readonly onServerOutput = (data: Buffer) => {
+        const output = data.toString().trim();
+
+        // Log filtering - keep for debugging
+        const suppressLevels = this.config.suppressLogLevels ?? ['INFO', 'DEBUG'];
```
satyakigh: This line can be moved out of the function, this only needs to execute once

tools/lspClient/LspClient.ts
```diff
@@ -0,0 +1,303 @@
...
+export class LspClient implements LspConnection {
+    protected serverProcess?: ChildProcess;
+    protected connection?: MessageConnection;
+
+    public readonly createdAt: number;
+    private readonly encryptionKey: Buffer;
+    protected isShutdown = false;
+    protected workspaceConfig: Record<string, unknown>[] = [{}];
+
+    constructor(protected config: LspClientConfig) {
```
satyakigh: protected?

tools/stability/Config.ts
```diff
@@ -0,0 +1,72 @@
+export interface Config {
+    duration: string;
+    maxRetries: number;
+    responseTimeout: number;
+    path: string;
+}
+
+function parseSimpleArgs(): Partial<Config> {
```
satyakigh, line 8: Use `yargs` similar to the other tools

tools/stability/Config.ts
```diff
@@ -0,0 +1,72 @@
+export interface Config {
```
satyakigh, line 1: This file should just return a `Config` object, instead of exporting the parsers

tools/stability/Templates.ts
```diff
@@ -0,0 +1,25 @@
+import { Templates } from '../../tst/utils/TemplateUtils';
+
+export interface TemplateConfig {
```
satyakigh: You can define this type in `TemplateUtils`  which uses 
```
{ fileName: string; contents: string }
```

tools/stability/Templates.ts
```diff
@@ -0,0 +1,25 @@
...
+    name: string;
+    content: string;
+}
+
+export const TEMPLATE_CONFIGS: TemplateConfig[] = [
```
satyakigh: Templates? Test Templates?
There is no config here

tools/stability/testers/TesterTypes.ts
```diff
@@ -0,0 +1,31 @@
...
+        maxDurationLimitMs: 3000,
+    },
+    [OperationType.COMPLETION]: {
+        retryTimeoutMs: 5000,
+        avgDurationLimitMs: 300,
+        maxDurationLimitMs: 5000,
+    },
+};
+
+export function getTesterConfig(operationType: OperationType): TesterConfig {
```
satyakigh: If `getTesterConfig` is required, we don't need to export `TESTER_CONFIG`, but easier would just be to export `TESTER_CONFIG` and use it directly `TESTER_CONFIG[operationType]` 

calling `getTesterConfig` or `TESTER_CONFIG[operationType]` once is trivial

tools/stability/testers/TesterUtils.ts
```diff
@@ -0,0 +1,27 @@
+import { recordOperation } from '../Monitoring';
```
satyakigh, line 1: Utils

.github/workflows/post-release-long-running.yml
```diff
@@ -0,0 +1,88 @@
...
+          else
+            unzip *.zip
+            chmod +x cfn-lsp-server-standalone.js
+          fi
+
+      - name: Test standalone bundle
+        run: node cfn-lsp-server-standalone.js --version
+
+      - name: Install dependencies
+        run: npm ci
```
satyakigh, line 80: This probably won't work for node 18

.github/workflows/post-beta-long-running.yml
```diff
@@ -0,0 +1,88 @@
...
+
+on:
+  release:
+    types: [published]
+
+permissions:
+  contents: read
+
+jobs:
+  check-beta-release:
```
satyakigh: this handles the main branch as well? Alpha?

.gitignore
chrisqm-dev: Removed !tools/**/*.ts

tools/lspClient/LspClient.ts
chrisqm-dev: Moved suppressLevels outside of the function

tools/lspClient/LspClient.ts
chrisqm-dev: Changed config to private

tools/stability/Config.ts
chrisqm-dev: `yargs` cannot be used in node 18: 



`Error: yargs parser supports a minimum Node.js version of 20. Read our version support policy: https://github.com/yargs/yargs-parser#supported-nodejs-versions`

tools/stability/Config.ts
chrisqm-dev: Fixed so that Config.ts exports a `config` object instead of `parseConfig()`

tools/stability/Templates.ts
chrisqm-dev: Renamed const to `TEST_TEMPLATES`

tools/stability/testers/TesterTypes.ts
chrisqm-dev: Removed getter function to instead use the const directly

tools/stability/testers/TesterUtils.ts
chrisqm-dev: Renamed file to "TesterUtils.ts"

.github/workflows/post-release-long-running.yml
chrisqm-dev: Confirmed this works for node 18 by running npm ci locally with node 18

.github/workflows/post-beta-long-running.yml
chrisqm-dev: Added a workflow for alpha to work. We need a release for this test to work as it requires a pre-built standalone file, which is not created on every push to main.

tools/stability/runStabilityTest.ts
```diff
@@ -0,0 +1,20 @@
+import { TestOrchestrator } from './TestOrchestrator';
+import { generateFinalReport } from './Monitoring';
+
+async function main(): Promise<void> {
+    const orchestrator = new TestOrchestrator();
+
+    try {
+        await orchestrator.initialize();
+        await orchestrator.runTests();
+        generateFinalReport(Date.now());
```
kddejong: `generateFinalReport(Date.now())` is called after the tests finish, so the reported runtime will always be ~0 minutes. Capture the start time before `runTests()` and pass that in. (Or just use the `startTime` that `initializeMonitoring()` already tracks internally.)

tools/lspClient/LspConnection.ts
```diff
@@ -0,0 +1,23 @@
+import { ClientInfo, AwsMetadata } from '../../src/server/InitParams';
+
+export interface LspConnection {
+    initialize(): Promise<void>;
+    sendRequest(method: string, params: any): Promise<any>;
+    sendNotification(method: string, params: any): Promise<void>;
```
kddejong, line 6: There's a lot of `any` throughout the client and testers — is there a reason we're not using the types from `vscode-languageserver-protocol`? Things like `Hover`, `CompletionList`, `CompletionParams`, etc. would give you compile-time safety on the validators for free.

tools/stability/Monitoring.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export function recordOperation(duration: number, operationType: OperationType): void {
+    const metric = metrics[operationType];
+    metric.operations++;
+
+    metric.averageDuration =
+        metric.averageDuration === null
+            ? duration
+            : (metric.averageDuration * (metric.operations - 1) + duration) / metric.operations;
+    metric.minDuration = metric.minDuration === null ? duration : Math.min(metric.minDuration, duration);
```
kddejong, line 32: `durations` is pushed to on every operation but never read anywhere. What's it for?

tools/stability/testers/HoverTester.ts
```diff
@@ -0,0 +1,89 @@
...
+Resources:
+  MyResource:
+    Type: AWS::S3::Bucket
+    Properties:
+      BucketName: TestName
+`;
+
+        await this.client.updateDocument(uri, 2, s3Template);
+
+        await retryOperationWithPerformance(
```
kddejong, line 46: Nit: the document versions (2, 3, 4, 5, 6) are hardcoded across the testers and orchestrator. Works fine now but could get fragile if more testers are added. A shared counter would help down the road.

kddejong (changes requested):
A few comments below, requesting changes for the report timing bug.

tools/stability/Monitoring.ts
chrisqm-dev: This was originally going to be used to generate metrics other than average (for example, median duration which requires the durations list). Those metrics were not implemented since average was sufficient. Removed the `durations` list, we can add it back if we later determine more metrics are needed

tools/lspClient/LspConnection.ts
chrisqm-dev: Added typing for the specific lsp requests. 

For sendRequest, sendNotification,  and onRequest,`any` will still be necessary since it can be used for any request object.

tools/stability/runStabilityTest.ts
chrisqm-dev: Fixed to use `startTime` value appropriately

tools/stability/testers/HoverTester.ts
chrisqm-dev: Added nextDocumentVersion() and resetDocumentVersion() helpers

kddejong (approved):
Ship it 🚀

rongyamazon (approved):

tools/stability/Templates.ts
chrisqm-dev: Created type

PR #536, opened 2026-04-23
Handle non-string parameter values in hover

chrisqm-dev:
*Description of changes:*
- If a parameter field (such as description) is not a string, `formatParameterHover` will throw a TypeError due to item.trim() requiring a string
- Added string type checking before item.trim() is called
  - Note: All parameter fields **must** be a string for a CloudFormation Template to be valid, CFN-Lint properly detects non-string fields and produces diagnostics for it 


src/hover/HoverFormatter.ts
```diff
@@ -1104,7 +1104,7 @@ export function formatParameterHover(parameter: Parameter): string {
         doc.push(`**Constraint Description:** ${parameter.ConstraintDescription}`);
     }
 
-    return doc.filter((item) => item.trim() !== '').join('\n\n');
+    return doc.filter((item) => typeof item === 'string' && item.trim() !== '').join('\n\n');
```
Zee2413: Can this be handled one level up in the call chain? The formatters should expect and work with strings. 

Maybe we can fix this more generally. I see more `.trim()` calls, for e.g., line 1125

chrisqm-dev: Agreed, fixed the string processing at `Entity.ts`. Instead of assuming type is `string | undefined`, the method determines if the field is a string and if not sets the field to undefined. Repeated this logic for all applicable fields on the Parameter

tst/unit/autocomplete/EntityFieldCompletionProvider.test.ts
```diff
@@ -42,7 +42,7 @@ describe('EntityFieldCompletionProvider', () => {
             const mockContext = createParameterContext('MyParameter', {
                 text: 'e',
                 data: {
-                    Type: 'string',
```
chrisqm-dev, line 45: lowercase `string` is not a valid type

Zee2413: Have a closer look at this. I believe this is document type.

kddejong (commented):
Thanks for the fix — the crash on non-string values is a real issue worth addressing.

One suggestion on the approach: instead of silently dropping non-string scalars to `undefined`, consider coercing them to strings. For example, if a user writes `Description: 12345` (a number) or `Description: true` (a boolean), the current PR would show no description at all in hover. Converting scalars to strings would preserve the user's intent — they'd see `"12345"` or `"true"` which is likely what they meant.

This is already the pattern used for `NoEcho` on the very next line: `String(object['NoEcho'])`.

A small reusable helper would keep this clean and could be used across the codebase wherever we parse template values:

```typescript
/** Convert scalar values (string, number, boolean) to string. Returns undefined for objects/arrays/undefined. */
function toStringOrUndefined(value: unknown): string | undefined {
    if (typeof value === 'string') return value;
    if (typeof value === 'number' || typeof value === 'boolean') return String(value);
    return undefined;
}
```

Then `Parameter.from` becomes:
```typescript
toStringOrUndefined(object['Description'])
toStringOrUndefined(object['AllowedPattern'])
toStringOrUndefined(object['ConstraintDescription'])
```

This way numeric/boolean descriptions render in hover instead of disappearing, objects (like intrinsic functions) are still correctly filtered out, and we have a utility that's reusable anywhere we need to safely extract string values from parsed template data.

src/context/semantic/Entity.ts
```diff
@@ -114,12 +114,14 @@ export class Parameter extends Entity {
 
         return new Parameter(
             logicalId,
-            object['Type'] as ParameterType | undefined,
+            typeof object['Type'] === 'string' && Object.values(ParameterType).includes(object['Type'] as ParameterType)
```
Zee2413: Use https://github.com/aws-cloudformation/cloudformation-languageserver/blob/main/src/context/semantic/ParameterType.ts#L62

tst/unit/context/semantic/Entity.test.ts
```diff
@@ -100,5 +100,39 @@ describe('Entity', () => {
             expect(parameter.name).toBe('TestParam');
             expect(parameter.NoEcho).toBe(true);
         });
+
+        it('should handle intrinsic functions in string fields by setting them to undefined', () => {
+            const data = {
+                Type: { 'Fn::Sub': 'String' }, // Object instead of string
```
Zee2413: Don't think we need comment at the end of every line of code here

tst/unit/autocomplete/EntityFieldCompletionProvider.test.ts
chrisqm-dev: createParameterContext() uses `Parameter.from()`. Since we're enforcing stricter validation on `Parameter.from()`, `Type: 'string'` is now getting rejected as the `ParameterType` enum only includes uppercase `String`. Updated this test to use `ParameterType.String` to keep the value consistent

tst/unit/context/semantic/Entity.test.ts
chrisqm-dev: Removed comments

chrisqm-dev:
> Thanks for the fix — the crash on non-string values is a real issue worth addressing.
> 
> One suggestion on the approach: instead of silently dropping non-string scalars to `undefined`, consider coercing them to strings. For example, if a user writes `Description: 12345` (a number) or `Description: true` (a boolean), the current PR would show no description at all in hover. Converting scalars to strings would preserve the user's intent — they'd see `"12345"` or `"true"` which is likely what they meant.
> 
> This is already the pattern used for `NoEcho` on the very next line: `String(object['NoEcho'])`.
> 
> A small reusable helper would keep this clean and could be used across the codebase wherever we parse template values:
> 
> ```ts
> /** Convert scalar values (string, number, boolean) to string. Returns undefined for objects/arrays/undefined. */
> function toStringOrUndefined(value: unknown): string | undefined {
>     if (typeof value === 'string') return value;
>     if (typeof value === 'number' || typeof value === 'boolean') return String(value);
>     return undefined;
> }
> ```
> 
> Then `Parameter.from` becomes:
> 
> ```ts
> toStringOrUndefined(object['Description'])
> toStringOrUndefined(object['AllowedPattern'])
> toStringOrUndefined(object['ConstraintDescription'])
> ```
> 
> This way numeric/boolean descriptions render in hover instead of disappearing, objects (like intrinsic functions) are still correctly filtered out, and we have a utility that's reusable anywhere we need to safely extract string values from parsed template data.

Created toStringOrUndefined method and updated Entity.ts

src/context/semantic/Entity.ts
chrisqm-dev: Modified to use the `PARAMETER_TYPES` const

Zee2413 (approved):

kddejong (approved):

PR #537, opened 2026-04-23
Filedb improv

satyakigh:
## Purpose

Prepare the CloudFormation Language Server to migrate its persistence layer from LMDB to an encrypted file-based datastore on all platforms (currently only Windows uses it). The migration is gated behind a new feature flag and the file store itself is hardened for production use.

---

### 1. New `FileDb` Feature Flag

A new feature flag controls whether the file-based datastore is used instead of LMDB on non-Windows systems.

| Environment | Enabled | Fleet % |
|-------------|---------|---------|
| Alpha       | Yes     | 100%    |
| Beta        | No      | 100%    |
| Prod        | No      | 0%      |

Only alpha is opted in 

### 2. EncryptedFileStore Hardening

The file-based datastore received significant reliability improvements across six areas:

- **Atomic writes with fsync** — After writing a temp file, data is flushed to disk before the atomic rename. The parent directory is also fsynced to ensure the directory entry is durable. Windows silently skips directory fsync since NTFS journaling handles durability.

- **Rename retry logic** — File renames now retry up to 10 times on transient OS errors (EPERM, EACCES, EBUSY, ENOENT). This handles Windows-specific issues like antivirus or indexers holding file handles. Failed temp files are cleaned up on final failure.

- **Serialized write queue** — Writes are chained through an internal promise queue so multiple async operations within the same process don't interleave. Each write waits for the previous one to finish before acquiring the cross-process file lock.

- **Improved locking strategy** — The lock target changed from the data file itself to the parent directory with an explicit lockfile path. This avoids issues where the data file gets replaced via rename while a lock still references it. Lock retry parameters were retuned: more retries, shorter intervals, and randomized jitter to reduce contention. The synchronous lock used during construction now retries instead of failing immediately.

- **Stale temp file cleanup** — On startup, the store scans its directory and removes leftover `.tmp` files from prior crashed processes. Only files matching the store's own name prefix are removed.

- **Resilience to external file deletion** — If the data file is deleted between operations (e.g., by another process), the store gracefully starts from an empty state instead of throwing an error.

---

### DIFF
--- a/src/datastore/file/EncryptedFileStore.ts
+++ b/src/datastore/file/EncryptedFileStore.ts
@@ -1,45 +1,79 @@
-import { existsSync, readFileSync, renameSync, statSync, writeFileSync } from 'fs'; // eslint-disable-line no-restricted-imports -- files being checked
-import { rename, writeFile } from 'fs/promises';
-import { join } from 'path';
+import {
+    closeSync,
+    existsSync,
+    fsyncSync,
+    openSync,
+    readdirSync,
+    readFileSync, // eslint-disable-line no-restricted-imports -- atomic file primitives required here
+    renameSync,
+    statSync,
+    unlinkSync,
+    writeFileSync,
+} from 'fs';
+import { open, rename, unlink, writeFile } from 'fs/promises';
+import { basename, join } from 'path';
 import { Logger } from 'pino';
 import { lock, LockOptions, lockSync } from 'proper-lockfile';
 import { LoggerFactory } from '../../telemetry/LoggerFactory';
 import { ScopedTelemetry } from '../../telemetry/ScopedTelemetry';
 import { TelemetryService } from '../../telemetry/TelemetryService';
+import { processId } from '../../utils/Environment';
+import { sleep } from '../../utils/Retry';
 import { DataStore } from '../DataStore';
 import { decrypt, encrypt } from './Encryption';
 
-const LOCK_OPTIONS_SYNC: LockOptions = { stale: 10_000 };
-const LOCK_OPTIONS: LockOptions = { ...LOCK_OPTIONS_SYNC, retries: { retries: 20, minTimeout: 50, maxTimeout: 1000 } };
+const STALE_MS = 30_000;
+const LOCK_OPTIONS_SYNC: LockOptions = { stale: STALE_MS, realpath: false };
+const LOCK_OPTIONS: LockOptions = {
+    ...LOCK_OPTIONS_SYNC,
+    retries: { retries: 100, minTimeout: 25, maxTimeout: 200, factor: 1.5, randomize: true },
+};
+
+const RENAME_MAX_RETRIES = 10;
+const RENAME_RETRY_DELAY_MS = 50;
+const RETRIABLE_RENAME_CODES = new Set(['EPERM', 'EACCES', 'EBUSY', 'ENOENT']);
+
+const SYNC_LOCK_MAX_ATTEMPTS = 200;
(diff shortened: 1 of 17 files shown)


tools/telemetry-generator.ts
```diff
@@ -193,7 +193,11 @@ function main() {
         stubInterface<LspSystemHandlers>(),
     );
 
-    const dataStoreFactory = new MultiDataStoreFactoryProvider();
+    const featureFlags = new FeatureFlagProvider(
+        getFromGitHub,
+        join(__dirname, '..', 'assets', 'featureFlag', `${AwsEnv}.json`),
```
kddejong, line 198: Seem to be some of this a lot.  `return join(__dirname, 'assets', 'featureFlag', `${AwsEnv.toLowerCase()}.json`);`.  utility?

kddejong: Can we use the one from `src/featureFlag/FeatureFlagProvider.ts`

src/datastore/file/EncryptedFileStore.ts
```diff
@@ -1,45 +1,79 @@
...
+import { processId } from '../../utils/Environment';
+import { sleep } from '../../utils/Retry';
 import { DataStore } from '../DataStore';
 import { decrypt, encrypt } from './Encryption';
 
-const LOCK_OPTIONS_SYNC: LockOptions = { stale: 10_000 };
-const LOCK_OPTIONS: LockOptions = { ...LOCK_OPTIONS_SYNC, retries: { retries: 20, minTimeout: 50, maxTimeout: 1000 } };
+const STALE_MS = 30_000;
+const LOCK_OPTIONS_SYNC: LockOptions = { stale: STALE_MS, realpath: false };
+const LOCK_OPTIONS: LockOptions = {
```
kddejong, line 27: nit: `cleanupStaleTmpFiles()` runs before the lock is acquired. Since tmp files are named `{name}.enc.{pid}.{counter}.tmp` and the cleanup matches any PID prefix for this store name, there's a small race with other language server processes (e.g. multiple VS Code windows sharing the same storageDir) — process A could delete a `.tmp` file that process B is about to `rename`.

Moving this call inside the lock block (after `lockSyncWithRetry`, before the `existsSync` check) would close that window.

src/datastore/file/EncryptedFileStore.ts
```diff
@@ -1,45 +1,79 @@
...
+            if (existsSync(this.file)) {
+                try {
+                    this.content = this.readFile();
+                } catch (error) {
+                    this.log.error(error, 'Failed to decrypt file store, recreating store');
+                    this.telemetry.count('filestore.recreate', 1);
+                    this.saveSync();
+                }
+            } else {
                 this.saveSync();
```
kddejong, line 73: `lockSyncWithRetry` does up to 200 × 50ms of `Atomics.wait`, which hard-blocks the Node.js event loop for up to 10 seconds. This runs in the constructor during LSP initialization, so the editor gets no responses while it's spinning.

The main scenario: process A crashes while holding the lock. The `stale` timeout is 30s, so process B retries synchronously until `proper-lockfile` considers the lock stale. If that takes longer than the 10s retry budget, the server fails to start.

This is an improvement over the previous no-retry `lockSync` (which would just fail immediately), but worth noting the tradeoff. Could the constructor fail fast and defer store initialization to an async path to avoid blocking?

tools/telemetry-generator.ts
satyakigh: The paths are different between the tools and src code, updated them to reuse

PR #548, opened 2026-04-27
Enable safeRestore config and retry op on recovery

satyakigh:
* Enable `safeRestore` configuration for LMDB - https://github.com/kriszyp/lmdb-js#readme 
* Retry operation if recovery was successful

### DIFF
--- a/src/datastore/LMDBStoreFactory.ts
+++ b/src/datastore/LMDBStoreFactory.ts
@@ -235,6 +235,10 @@ export class LMDBStoreFactory implements DataStoreFactory {
         if (this.closed) return;
 
         try {
+            const staleLocks = this.env.readerCheck();
+            if (staleLocks > 0) {
+                this.log.info(`Removed ${staleLocks} stale reader locks for LMDB`);
+            }
             const envStat = stats(this.env);
             this.telemetry.histogram('version', VersionNumber);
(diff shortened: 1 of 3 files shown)


src/datastore/LMDBStoreFactory.ts
```diff
@@ -235,6 +235,10 @@ export class LMDBStoreFactory implements DataStoreFactory {
         if (this.closed) return;
 
         try {
+            const staleLocks = this.env.readerCheck();
```
Zee2413, line 238: how does this work?

Zee2413 (approved):

src/datastore/LMDBStoreFactory.ts
satyakigh: **readerCheck** clears out stale reader transaction entries from the reader lock table.

LMDB uses a shared lock table to track active read transactions across processes/threads. If a process crashes or exits without properly closing its read transaction, its entry remains in the table as a stale reader. Stale readers are problematic because LMDB can't reclaim database pages that were in use at the time of that reader's snapshot - this causes the database file to grow continuously since old pages can never be freed.

db.readerCheck() scans the reader lock table, detects entries belonging to processes/PIDs that no longer exist, and clears them. It returns the number of stale entries that were removed.

PR #552, opened 2026-05-01
Create a local file handler for concurrent read/writes

satyakigh:
Multiple language server processes (e.g., two IDE windows) share the same on-disk storage directory. `EncryptedFile` had inline locking via `proper-lockfile` and wrote through temp-file-then-rename, but `FeatureFlagProvider` called `writeFileSync` directly with no locking or atomicity. Both paths lacked fsync, so data could be lost on failure. Crashed processes left orphaned `.tmp` files that were never cleaned up.

## Changes
A new `LocalFile` class wraps a file path and provides read, write, and delete operations. All writes follow this protocol:

1. Acquire an inter-process lock using 
2. Scan the parent directory and delete any stale `.tmp` files matching `{fileName}.*.tmp` left by previous crashed writes
3. Write data to `{file}.{pid}.{counter}.tmp` — the PID + monotonic counter ensure uniqueness
4. `fsync` the temp file (flush kernel buffers to disk)
5. `rename` temp → target with retry on transient Windows errors (`EPERM`, `EACCES`, `EBUSY`, `ENOENT` — up to 10 retries, 50ms apart)
6. `fsync` the parent directory (makes the rename durable; silently skipped on Windows which doesn't support dir fsync)
7. Release the lock
8. If any step 3–6 fails: delete the temp file, then re-throw

Reads are lockless - safe because `rename` is atomic on POSIX, so readers always see a complete file.

Deletes acquire the lock, call `unlink`, and treat `ENOENT` as success (idempotent). `removeSync` is the exception — it skips locking, used only for corrupt-file cleanup in the `EncryptedFile` constructor.

### DIFF
--- a/src/utils/File.ts
+++ b/src/utils/File.ts
@@ -1,6 +1,9 @@
 import { readFileSync, existsSync, PathLike } from 'fs'; // eslint-disable-line no-restricted-imports
-import { readFile } from 'fs/promises'; // eslint-disable-line no-restricted-imports
+import { readFile, rename, unlink, open } from 'fs/promises'; // eslint-disable-line no-restricted-imports
 import { LoggerFactory } from '../telemetry/LoggerFactory';
+import { isWindows } from './Environment';
+import { DoesNotExist } from './Errors';
+import { sleep } from './Retry';
 import { toString } from './String';
 
 const log = LoggerFactory.getLogger('File');
@@ -17,7 +20,7 @@ export function readFileIfExists(path: PathLike, options: Options = 'utf8'): str
         if (existsSync(path)) {
             return readFileSync(path, options);
         } else {
-            throw new Error(`Path ${toString(path)} does not exist`);
+            throw new DoesNotExist(toString(path));
         }
     } catch (err) {
         log.error(err);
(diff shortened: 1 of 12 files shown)


src/utils/File.ts
```diff
@@ -49,10 +52,88 @@ export function readBufferIfExists(
...
         } else {
-            throw new Error(`Path ${toString(path)} does not exist`);
+            throw new DoesNotExist(toString(path));
         }
     } catch (err) {
         log.error(err);
         throw err;
     }
 }
+
```
kddejong, line 62: Nit: `existsSync` then `readFile` has a small TOCTOU gap — the file could disappear between the two calls. The existing read functions have the same pattern so not new, but since you're adding a fresh one, might be worth just calling `readFile` and catching `ENOENT`.

src/utils/LocalFile.ts
```diff
@@ -0,0 +1,217 @@
...
+            await this.cleanupStaleTmpFiles();
+            const tmp = this.tmpPath();
+
+            try {
+                await writeFile(tmp, data);
+                await asyncFileSync(tmp);
+                await asyncRenameWithRetry(tmp, this.path);
+
+                await asyncDirSync(dirname(this.path));
+                return true;
```
kddejong, line 131: Lockless reads work because `rename` is atomic on POSIX, but on Windows it's effectively delete+create, so a reader could hit a transient failure during a concurrent write. Probably fine here since callers already handle read errors, but a short comment explaining the tradeoff would help future readers.

kddejong (approved):
Nice cleanup. The write protocol looks correct and the new `LocalFile` abstraction simplifies `EncryptedFile` a lot. Two minor comments below.

PR #560, opened 2026-05-05
fix: add close-guard to prevent TOCTOU race in LMDB lifecycle

gemammercado:
Add a close-guard to synchronize database operations with environment lifecycle (close/reopenEnv). This prevents the race condition where close() or reopenEnv() can execute while async operations are still in-flight, causing 'Can not renew a transaction from a closed database' errors.

1. First `LMDBStore.execAsync()` calls `validateDatabase()` -> this passes -> then `await fn()` yields to the event loop
2. While yielded, `LMDBStoreFactory.close()` or `reopenEnv()` runs, sets `closed = true`, calls `env.close()` which aborts all active transactions and nulls the native LMDB handle
3. The awaited `fn()` resumes and tries `mdb_txn_renew()` on the now-closed environment

This approach uses an activeOps counter instead of locks:
- beginOp() — called at the start of every operation. Rejects with "Database is closed" if shutdown has started, otherwise increments the counter and returns a release function.
- close() — sets closed = true to reject new operations, then polls until activeOps === 0 (max 2s) before closing the env. Logs a warning if the timeout is hit.
    reopenEnv() — no draining needed. It's synchronous, and execAsync's retry logic handles any in-flight ops that fail on the old env.

Safe without locks because JS is single-threaded: the closed check + counter increment in beginOp() cannot be interrupted.

*Issue #, if available:*
This fix addresses provider latency spikes.
*Description of changes:*

### DIFF
--- a/src/datastore/LMDBStoreFactory.ts
+++ b/src/datastore/LMDBStoreFactory.ts
@@ -23,6 +23,7 @@ export class LMDBStoreFactory implements DataStoreFactory {
     private env: RootDatabase;
     private openPid = processId();
     private closed = false;
+    private activeOps = 0;
 
     private readonly stores = new Map<StoreName, LMDBStore>();
 
@@ -96,10 +97,29 @@ export class LMDBStoreFactory implements DataStoreFactory {
         if (this.closed) return;
         this.closed = true;
 
+        const deadline = Date.now() + 2000;
+        while (this.activeOps > 0 && Date.now() < deadline) {
+            await new Promise((resolve) => setTimeout(resolve, 5));
+        }
+        if (this.activeOps > 0) {
+            this.log.warn({ activeOps: this.activeOps }, 'Closing LMDB with in-flight operations after timeout');
+        }
+
         clearInterval(this.metricsInterval);
         clearTimeout(this.timeout);
         this.stores.clear();
         await this.env.close();
+    }
+
+    private beginOp(): () => void {
+        // Safe: JS is single-threaded, so closed check + increment is atomic within a tick
+        if (this.closed) {
+            throw new Error('Database is closed');
+        }
+        this.activeOps++;
+        return () => {
+            this.activeOps--;
+        };
     }
 
     private handleError(error: unknown): void {
@@ -175,6 +195,12 @@ export class LMDBStoreFactory implements DataStoreFactory {
         }
     }
 
+    /**
+     * Replaces the LMDB environment synchronously. In-flight async operations may
+     * fail on the old env, but execAsync's retry logic will recover using the new
+     * store handle (updated by recreateStores). Unlike close(), we don't drain
+     * activeOps because the env is being replaced, not destroyed.
+     */
     private reopenEnv(): void {
         this.telemetry.count('env.reopen', 1);
         this.env = createEnv(this.lmdbDir).env;
@@ -202,6 +228,7 @@ export class LMDBStoreFactory implements DataStoreFactory {
                 database,
                 (e) => this.handleError(e),
                 () => this.ensureValidEnv(),
+                () => this.beginOp(),
             ),
         );
     }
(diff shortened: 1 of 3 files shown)


src/datastore/LMDBStoreFactory.ts
```diff
@@ -175,6 +182,18 @@ export class LMDBStoreFactory implements DataStoreFactory {
         }
     }
 
+    private async reopenEnvExclusive(): Promise<void> {
```
kddejong: I don't see this code being called? Is it needed?

src/datastore/LMDBStoreFactory.ts
```diff
@@ -94,12 +96,17 @@ export class LMDBStoreFactory implements DataStoreFactory {
 
     async close(): Promise<void> {
```
kddejong, line 96: close() still has a race on the this.closed early-return check

Two concurrent close() calls could both pass the if (this.closed) check before either acquires the write lock. The second would then try to close an already-closed env.

Move the check inside the lock:
```
async close(): Promise<void> {
    const release = await this.envLock.writeLock();
    try {
        if (this.closed) return;
        this.closed = true;
        // ...
    } finally {
        release();
    }
}
```

kddejong (changes requested):

src/datastore/LMDBStoreFactory.ts
gemammercado: leftover, removed

src/datastore/LMDBStoreFactory.ts
```diff
@@ -93,13 +95,17 @@ export class LMDBStoreFactory implements DataStoreFactory {
...
-        if (this.closed) return;
-        this.closed = true;
-
-        clearInterval(this.metricsInterval);
-        clearTimeout(this.timeout);
-        this.stores.clear();
-        await this.env.close();
+        const release = await this.envLock.writeLock();
```
satyakigh: If it's already closed, we don't need the lock

src/datastore/lmdb/LMDBStore.ts
```diff
@@ -41,21 +43,26 @@ export class LMDBStore implements DataStore {
...
-                } catch (e) {
-                    this.onError(e);
-                    this.telemetry.count(`retry.${op}`, 1);
-                    this.validateDatabase();
-                    return await fn();
-                }
-            },
-            { captureErrorAttributes: true },
-        );
+        const release = await this.envLock.readLock();
```
satyakigh: Lock is part of latency it should be inside measure

src/datastore/lmdb/AsyncRWLock.ts
```diff
@@ -0,0 +1,54 @@
...
+ * - Write lock: acquired by close()/reopenEnv() (exclusive, waits for all reads to drain)
+ */
+export class AsyncRWLock {
+    private readers = 0;
+    private writer = false;
+    private readonly waitingWriters: Array<() => void> = [];
+    private readonly waitingReaders: Array<() => void> = [];
+
+    async readLock(): Promise<() => void> {
+        while (this.writer || this.waitingWriters.length > 0) {
```
satyakigh: Can this indefinitely run?

src/datastore/LMDBStoreFactory.ts
gemammercado: The lock is there to prevent a close when any read ops are in flight. It's not there for the if check, but it is convenient to place it after the lock to avoid any concurrent closes. A double close() would throw the same error we are trying to avoid 'Can not renew a transaction from a closed database.'

src/datastore/lmdb/AsyncRWLock.ts
gemammercado: No, acquiring the write lock is wrapped in a try-finally block so it always releases even when the close fails

gemammercado (commented):
responded to @satyakigh feedback

kddejong (commented):
Posted in error, please disregard.

kddejong (approved):
LGTM

satyakigh (approved):

PR #563, opened 2026-05-05
fix: filtering out client errors for all AWS API operations decorated with @Measur…

vzhan00:
…e and @Count at the caller level.

*Description of changes:*

For our online components we need to prevent client errors from emitting fault metrics. To prevent user experience from changing while still ensuring only AWS calls have client error faults not being emitted, at the service level we add a flag to the error thrown to indicate to our decorators that faults should not be emitted if they are AWS client errors.

### DIFF
--- a/src/services/CcapiService.ts
+++ b/src/services/CcapiService.ts
@@ -7,8 +7,7 @@ import {
 } from '@aws-sdk/client-cloudcontrol';
 import { LoggerFactory } from '../telemetry/LoggerFactory';
 import { Measure } from '../telemetry/TelemetryDecorator';
-import { isClientError } from '../utils/AwsErrorMapper';
-import { markSuppressFault } from '../utils/FaultSuppression';
+import { markIfClientError } from '../utils/FaultSuppression';
 import { AwsClient } from './AwsClient';
 
 const log = LoggerFactory.getLogger('CcapiService');
@@ -27,9 +26,7 @@ export class CcapiService {
             return await request(client);
         } catch (error) {
             log.error(error, 'CloudControl API call failed');
-            if (error instanceof Error && isClientError(error)) {
-                markSuppressFault(error);
-            }
+            markIfClientError(error);
             throw error;
         }
     }
--- a/src/services/CcapiService.ts
+++ b/src/services/CcapiService.ts
@@ -7,7 +7,8 @@ import {
 } from '@aws-sdk/client-cloudcontrol';
 import { LoggerFactory } from '../telemetry/LoggerFactory';
 import { Measure } from '../telemetry/TelemetryDecorator';
-import { mapAwsErrorToLspError } from '../utils/AwsErrorMapper';
+import { isClientError } from '../utils/AwsErrorMapper';
+import { markSuppressFault } from '../utils/FaultSuppression';
 import { AwsClient } from './AwsClient';
 
 const log = LoggerFactory.getLogger('CcapiService');
@@ -26,7 +27,10 @@ export class CcapiService {
             return await request(client);
         } catch (error) {
             log.error(error, 'CloudControl API call failed');
-            throw mapAwsErrorToLspError(error, { isAwsCall: true });
+            if (error instanceof Error && isClientError(error)) {
+                markSuppressFault(error);
+            }
+            throw error;
         }
     }
 
--- a/src/services/CcapiService.ts
+++ b/src/services/CcapiService.ts
@@ -5,8 +5,12 @@ import {
     ListResourcesCommand,
     ListResourcesOutput,
 } from '@aws-sdk/client-cloudcontrol';
+import { LoggerFactory } from '../telemetry/LoggerFactory';
 import { Measure } from '../telemetry/TelemetryDecorator';
+import { mapAwsErrorToLspError } from '../utils/AwsErrorMapper';
 import { AwsClient } from './AwsClient';
+
+const log = LoggerFactory.getLogger('CcapiService');
 
 export interface ListResourcesOptions {
     nextToken?: string;
@@ -17,8 +21,13 @@ export class CcapiService {
     constructor(private readonly awsClient: AwsClient) {}
 
     private async withClient<T>(request: (client: CloudControlClient) => Promise<T>): Promise<T> {
-        const client = this.awsClient.getCloudControlClient();
-        return await request(client);
+        try {
+            const client = this.awsClient.getCloudControlClient();
+            return await request(client);
+        } catch (error) {
+            log.error(error, 'CloudControl API call failed');
+            throw mapAwsErrorToLspError(error, { isAwsCall: true });
+        }
     }
 
     @Measure({ name: 'listResources' })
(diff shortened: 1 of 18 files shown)


src/telemetry/ScopedTelemetry.ts
```diff
@@ -109,7 +110,9 @@ export class ScopedTelemetry implements Closeable {
         try {
             return fn();
         } catch (error) {
-            this.error(`${name}.fault`, error, undefined, config);
+            if (!isClientError(error)) {
```
satyakigh: Adding it at this level will drop fault metrics from valid non AWS api calls like axios, database, etc. `isClientError` needs to be added at the AWS API call level

vzhan00: Wrapped at the AWS API caller level to catch, log, and return

chrisqm-dev:
Do we need to filter out client errors on the artifact exporter as well? https://github.com/aws-cloudformation/cloudformation-languageserver/tree/main/src/artifactexporter

vzhan00:
Good call-out, added filter to artifact explorer

src/artifactexporter/ArtifactExporter.ts
```diff
@@ -130,7 +134,15 @@ export class ArtifactExporter {
             return this.templateDict;
         }
 
-        await this.exportResources(bucketName, s3KeyPrefix);
+        try {
+            await this.exportResources(bucketName, s3KeyPrefix);
+        } catch (error) {
+            if (isClientError(error)) {
+                log.warn(error, 'Client error during artifact export');
```
chrisqm-dev: can we make tests for each client error case?

vzhan00: Added tests

Zee2413:
Looks like a lot of errors are being swallow that were designed to be surfaced. We built a handler to inform the customer of the errors. https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/awsService/cloudformation/utils/onlineErrorHandler.ts

Why are we making this change? To omit client errors from contributing to fault metrics? If so we should find a way to do that without changing the overall customer experience / behavior.

vzhan00:
With every online call wrapped by the [OnlineFeatureWrapper](https://github.com/aws-cloudformation/cloudformation-languageserver/blob/main/src/utils/OnlineFeatureWrapper.ts) that leverages `mapAwsErrorToLspError`. Instead of using `isClientError`, to address both concerns: https://github.com/aws-cloudformation/cloudformation-languageserver/pull/563#issuecomment-4423283944 and https://github.com/aws-cloudformation/cloudformation-languageserver/pull/563#discussion_r3191532578, we can add a flag to [OnlineFeatureErrorData](https://github.com/aws-cloudformation/cloudformation-languageserver/blob/main/src/utils/OnlineFeatureError.ts#L10)  and within `mapAwsErrorToLspError` add that flag to the `ResponseError` thrown. We can then check for this flag in [ScopedTelemetry](https://github.com/aws-cloudformation/cloudformation-languageserver/blob/main/src/telemetry/ScopedTelemetry.ts#L112) when handling errors and not emit the fault metric there. This way we only set this flag when we want to at the AWS API call level while still surfacing error popups for users.

Zee2413:
> With every online call wrapped by the [OnlineFeatureWrapper](https://github.com/aws-cloudformation/cloudformation-languageserver/blob/main/src/utils/OnlineFeatureWrapper.ts) that leverages `mapAwsErrorToLspError`. Instead of using `isClientError`, to address both concerns: [#563 (comment)](https://github.com/aws-cloudformation/cloudformation-languageserver/pull/563#issuecomment-4423283944) and [#563 (comment)](https://github.com/aws-cloudformation/cloudformation-languageserver/pull/563#discussion_r3191532578), we can add a flag to [OnlineFeatureErrorData](https://github.com/aws-cloudformation/cloudformation-languageserver/blob/main/src/utils/OnlineFeatureError.ts#L10) and within `mapAwsErrorToLspError` add that flag to the `ResponseError` thrown. We can then check for this flag in [ScopedTelemetry](https://github.com/aws-cloudformation/cloudformation-languageserver/blob/main/src/telemetry/ScopedTelemetry.ts#L112) when handling errors and not emit the fault metric there. This way we only set this flag when we want to at the AWS API call level while still surfacing error popups for users.

Sounds like this will work to filter the fault metrics and preserve the client behavior/error handling.

src/services/CcapiService.ts
```diff
@@ -17,8 +21,13 @@ export class CcapiService {
...
 
     private async withClient<T>(request: (client: CloudControlClient) => Promise<T>): Promise<T> {
-        const client = this.awsClient.getCloudControlClient();
-        return await request(client);
+        try {
+            const client = this.awsClient.getCloudControlClient();
+            return await request(client);
+        } catch (error) {
+            log.error(error, 'CloudControl API call failed');
+            throw mapAwsErrorToLspError(error, { isAwsCall: true });
```
Zee2413: Throwing from the `withClient` method of these four services turns any and all exception to `ResponseError` type masking the original exception type. Any upstream caller checking for service specifics exceptions will never catch the expected types.

One example: https://github.com/aws-cloudformation/cloudformation-languageserver/blob/main/src/resourceState/ResourceStateManager.ts#L80

Now this caller will never see a `ResourceNotFoundException` which is a common CCAPI exception type. It is also dead code now.

vzhan00: Marking the error with union types and rethrowing instead of changing the error type now.

src/telemetry/ScopedTelemetry.ts
```diff
@@ -109,7 +110,9 @@ export class ScopedTelemetry implements Closeable {
         try {
             return fn();
         } catch (error) {
-            this.error(`${name}.fault`, error, undefined, config);
+            if (!hasSuppressFault(error)) {
+                this.error(`${name}.fault`, error, undefined, config);
```
satyakigh, line 117: Instead of dropping the metric entirely, can publish a `.error` instead

src/utils/FaultSuppression.ts
```diff
@@ -0,0 +1,11 @@
+export interface Suppressible {
```
satyakigh, line 3: Can this be a custom Error class instead? Wrap the error into a new class that we control with `suppress` or other fields as attributes of the class

src/services/S3Service.ts
```diff
@@ -1,15 +1,28 @@
...
 
     protected async withClient<T>(request: (client: S3Client) => Promise<T>): Promise<T> {
-        const client = this.awsClient.getS3Client();
-        return await request(client);
+        try {
+            const client = this.awsClient.getS3Client();
+            return await request(client);
+        } catch (error) {
+            log.error(error, 'S3 API call failed');
+            if (error instanceof Error && isClientError(error)) {
```
satyakigh: This can probably be simplified and unified - a single function that:
1. Check its an error
2. Checks whether its an AWS 4xx failure
3. Converts to the "new" error type or otherwise

src/utils/FaultSuppression.ts
Zee2413: This can be done but callers of ccapi/cfn/s3 service should be expecting service specific exception types, not a custom one. Example https://github.com/aws-cloudformation/cloudformation-languageserver/blob/main/src/resourceState/ResourceStateManager.ts#L80

We'll have to update all callers / catch cases.

vzhan00: As @Zee2413 said, we need to preserve error types in this case so those lines don't end up becoming dead code

Zee2413 (approved):

satyakigh (approved):

PR #586, opened 2026-05-20
Use s3 to verify bucket ownership during get resource state

Zee2413:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/services/S3Service.ts
+++ b/src/services/S3Service.ts
@@ -10,7 +10,6 @@ import {
 } from '@aws-sdk/client-s3';
 import { LoggerFactory } from '../telemetry/LoggerFactory';
 import { Measure } from '../telemetry/TelemetryDecorator';
-import { isClientError } from '../utils/AwsErrorMapper';
 import { markIfClientError } from '../utils/FaultSuppression';
 import { AwsClient } from './AwsClient';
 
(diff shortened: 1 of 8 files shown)


src/services/S3Service.ts
```diff
@@ -64,6 +72,34 @@ export class S3Service {
...
     }
 
+    @Measure({ name: 'verifyBucketAccessibleInRegion' })
+    async verifyBucketAccessibleInRegion(bucketName: string, region: string): Promise<string | undefined> {
+        const client = this.awsClient.getS3Client();
+
+        try {
+            await client.send(new GetBucketEncryptionCommand({ Bucket: bucketName }));
+        } catch (error) {
+            if (isClientError(error)) {
```
chrisqm-dev: isClientError also contains other categories like network issues. Should we create a separate error parsing method for this use case? same with the other isClientError call

Zee2413: @chrisqm-dev Thanks for catching this — your concern was correct, and following the thread led to a couple of substantive design changes. Pushed as two follow-up commits.

**What `isClientError` was actually doing**

You were right that `isClientError` collapses several categories (credentials, network, permissions, 4xx service). The first follow-up commit just lets exceptions propagate and removes the invented "not owned" / "not accessible" strings. The strings were misleading: a network timeout or expired credential would be reported to the user as "bucket is not owned by the current account," which sends them down the wrong debugging path.

**Why the verification step exists at all (re-derived empirically)**

I wanted to confirm the verification step was necessary before reworking it. Tested directly against the public `commoncrawl` bucket with my own credentials:

```
$ aws cloudcontrol get-resource --type-name AWS::S3::Bucket --identifier commoncrawl ...
{ "Properties": "..." }   # ← full bucket properties returned
```

CCAPI's S3 read handler returns full property data for any bucket the caller can read, not just buckets they own. So a verification step is genuinely needed to prevent a user from accidentally pulling info about a third-party public bucket they typed by mistake.

**Why `HeadBucket` alone isn't sufficient**

Tested every candidate API against four scenarios. Key result:

| | `HeadBucket` | `HeadBucket` + `ExpectedBucketOwner` |
|---|---|---|
| `commoncrawl` (public, cross-account) | ✓ 200 (leak) | ✗ 403 |
| Bucket I own (any region) | ✓ 200 + `BucketRegion` | ✓ 200 + `BucketRegion` |
| Bucket genuinely not mine | ✗ 403 | ✗ 403 |
| Non-existent | ✗ 404 | ✗ 404 |

Without `ExpectedBucketOwner`, `HeadBucket` 200s for publicly readable cross-account buckets — exactly the case the verification is supposed to prevent. With `ExpectedBucketOwner` it correctly 403s.

**Final shape (commit d2f4b92)**

Replaced `GetBucketEncryption + HeadBucket` with a single `HeadBucket(ExpectedBucketOwner=<caller account>)` call:

- Caller account resolved via `sts:GetCallerIdentity` (no IAM permission required — always available to authenticated principals).
- 403 / 404 → throw `ResourceNotFoundException`. From the user's perspective "doesn't exist" and "you don't own it" are the same outcome, and translating to the existing CCAPI exception lets `ResourceStateManager.getResource` handle it through the same code path as any other not-found result.
- Wrong region → distinct error string (kept as-is — it's a user-actionable, recoverable signal).
- Network / credential / 5xx errors → propagate unchanged with the real SDK message.

**Permissions**

`HeadBucket` requires `s3:ListBucket`, which is already in the `AWS::S3::Bucket` read handler permission set. No new permission burden — anyone with permission to use CCAPI on this resource type already has it.

**Tests**

- Unit tests rewritten for the new behavior, including STS failure modes and the four error-path categories.
- Three new e2e tests assert that CCAPI is NOT called when verification fails (cross-account 403, missing bucket 404, wrong region).
- Shared mock infra (`MockAwsTestClient`) extended with `mockStsSend` so future S3-related e2e tests don't have to reinvent it.

chrisqm-dev (approved):

vzhan00 (approved):

PR #588, opened 2026-05-21
Adding support for VSCode command suffix's per extension

vzhan00:
**Description of changes:**
  
  Fixes duplicate command registration crash when multiple VS Code extensions (e.g., AWS Toolkit and CloudFormation standalone) run the same language server simultaneously.
  
  Problem: vscode-languageclient registers commands from executeCommandProvider in VS Code's global command registry, which doesn't allow duplicates. The second extension to activate crashes with Error: command '/command/template/clear-diagnostic' already exists.
  
  Fix: The server now derives a unique command suffix from initializationOptions.aws.clientInfo.extension.name (already sent by all clients). Each client gets uniquely-suffixed commands (e.g., /command/template/clear-diagnostic.aws.cloudformation), eliminating the collision.
  
  **Changes:**
  
  - ExecutionHandler.ts — createCommands(suffix) factory, module-level singleton with initCommands()/getCommands()
  - LspCapabilities.ts — converted from static const to buildCapabilities(commands) function
  - CodeActionService.ts — replaced hardcoded command strings with getCommands() calls
  - standalone.ts — reads extension name from init params, calls initCommands()
  - LspConnection.ts — updated default handler to use buildCapabilities()
  
  Backward compatible: If no extension name is provided, commands remain unsuffixed (original behavior). Currently since we're using the extension name, the existing toolkit suffix will be `amazonwebservices.aws-toolkit-vscode` and for the standalone it will be `aws.cloudformation`.
  
  **Testing:**
  
  - 3328 unit tests pass
  - Manually verified both Toolkit and standalone extension running simultaneously without crash
  - Verified code actions fire correctly with suffixed commands

  
### DIFF
--- a/src/app/standalone.ts
+++ b/src/app/standalone.ts
@@ -1,7 +1,8 @@
 import './polyfills';
 import { createConnection, ProposedFeatures } from 'vscode-languageserver/node'; // eslint-disable-line no-restricted-imports
 import { InitializedParams } from 'vscode-languageserver-protocol';
-import { LspCapabilities } from '../protocol/LspCapabilities';
+import { initCommands, getCommands } from '../handlers/ExecutionHandler';
+import { buildCapabilities } from '../protocol/LspCapabilities';
 import { LspConnection } from '../protocol/LspConnection';
 import { ExtendedInitializeParams } from '../server/InitParams';
 import { ExtensionName } from '../utils/ExtensionConfig';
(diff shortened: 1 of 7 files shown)


src/app/standalone.ts
```diff
@@ -13,13 +14,17 @@ let server: unknown;
 async function onInitialize(params: ExtendedInitializeParams) {
     staticInitialize(params.clientInfo, params.initializationOptions?.['aws']);
 
+    const name = params.initializationOptions?.aws?.clientInfo?.extension?.name;
+    const suffix = typeof name === 'string' ? name : undefined;
```
satyakigh, line 18: We should only allow this if its a string, otherwise throw

src/handlers/ExecutionHandler.ts
```diff
@@ -4,6 +4,34 @@ import { LoggerFactory } from '../telemetry/LoggerFactory';
 import { TelemetryService } from '../telemetry/TelemetryService';
 import { getRegion } from '../utils/Region';
 
+export const CLEAR_DIAGNOSTIC = '/command/template/clear-diagnostic';
```
satyakigh, line 7: Instead of this singleton pattern, can the suffix be used to directly create LspCapabilities + LspConnection. Then just send the capabilities which has the set of commands directly to `CfnServer` creation and pass the commands down to execution handler and code action handler?

PR #598, opened 2026-06-03
fix: classify missing credentials as a client error

mrinaudo-aws:
*Issue #, if available:*

*Description of changes:*
The current code today catches a missing credential-related error, and rethrows it as a new one with a generic name (`Error`), that is not being recognized as a credential-related issue.  The fix, proposed in this PR, lets the original error to propagate untouched.

### DIFF
--- a/src/auth/AwsCredentials.ts
+++ b/src/auth/AwsCredentials.ts
@@ -10,6 +10,7 @@ import { Telemetry } from '../telemetry/TelemetryDecorator';
 import { getRegion } from '../utils/Region';
 import { parseWithPrettyError } from '../utils/ZodErrorWrapper';
 import { UpdateCredentialsParams, IamCredentials } from './AwsLspAuthTypes';
+import { CredentialsProviderError } from './CredentialsProviderError';
 
 const DecryptedCredentialsSchema = z.object({
     data: z.object({
@@ -50,9 +51,7 @@ export class AwsCredentials {
 
     getIAM(): DeepReadonly<IamCredentials> {
         if (!this.iamCredentials) {
-            const error = new Error('IAM credentials not configured');
-            error.name = 'CredentialsProviderError';
-            throw error;
+            throw new CredentialsProviderError();
         }
         return structuredClone(this.iamCredentials);
     }
--- a/src/auth/AwsCredentials.ts
+++ b/src/auth/AwsCredentials.ts
@@ -51,7 +51,7 @@ export class AwsCredentials {
 
     getIAM(): DeepReadonly<IamCredentials> {
         if (!this.iamCredentials) {
-            throw new CredentialsProviderError();
+            throw new CredentialsProviderError('IAM credentials not configured');
         }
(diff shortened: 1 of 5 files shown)


Zee2413 (dismissed):

src/auth/AwsCredentials.ts
```diff
@@ -50,7 +50,9 @@ export class AwsCredentials {
 
     getIAM(): DeepReadonly<IamCredentials> {
         if (!this.iamCredentials) {
-            throw new Error('IAM credentials not configured');
+            const error = new Error('IAM credentials not configured');
+            error.name = 'CredentialsProviderError';
```
satyakigh: Lets create a fully qualified class instead of just renaming the `name`

mrinaudo-aws: Thank you, @satyakigh !  Let me look into it!

src/auth/CredentialsProviderError.ts
```diff
@@ -0,0 +1,7 @@
+export class CredentialsProviderError extends Error {
```
Zee2413, line 1: class name seems too generic as compared to the message 'not configured'.

mrinaudo-aws: Thank you, @Zee2413 - `CredentialsProviderError` would be the same name the AWS SDK uses for this class of error, so perhaps leaving it as it is should be reasonably safe.

src/auth/CredentialsProviderError.ts
```diff
@@ -0,0 +1,7 @@
...
+    constructor(message: string = 'IAM credentials not configured') {
+        super(message);
+        this.name = 'CredentialsProviderError';
+        Object.setPrototypeOf(this, CredentialsProviderError.prototype);
```
Zee2413: Don't think `Object.setPrototypeOf` is necessary.

src/auth/CredentialsProviderError.ts
Zee2413: Sounds good. The message can be added when instantiating rather than be the default hardcoded string.

Zee2413 (dismissed):

src/auth/CredentialsProviderError.ts
mrinaudo-aws: Let me look into this, thank you!

src/auth/CredentialsProviderError.ts
mrinaudo-aws: Let me look into this, thank you!

src/auth/CredentialsProviderError.ts
mrinaudo-aws: I initially expunged it in a commit subsequent to me reading this feedback, but I just readded it back in a new commit, for consistency and for `instanceof` safety.

satyakigh (approved):

kddejong (approved):

Zee2413 (approved):

PR #599, opened 2026-06-04
Do not call find occurences multiple times

satyakigh:
`generateCodeActions` is synchronous, so its measured duration is main-thread block time. The "Extract to Parameter" refactor ran the full Resources+Outputs literal scan (AllOccurrencesFinder.findAllOccurrences) twice per request - once in `hasMultipleOccurrences`, then again in `generateAllOccurrencesExtraction`. 

From metric for the worst real sample: total 12,003 ms = hasMultipleOccurrences 6,145 ms + extractAllToParameter 5,978 ms. While the actual extract action was only 60 ms. Because the editor requests code actions on cursor moves and diagnostic refreshes, this caused repeated multi-second freezes.

1. CodeActionService.generateRefactorActions - removed the separate hasMultipleOccurrences scan; the extract-all path now runs the scan only once.
2. ExtractToParameterProvider.generateAllOccurrencesExtraction - returns undefined when occurrences ≤ 1, preserving the prior "offer Extract-All only for multiple occurrences" behavior now that the gating scan is gone.
3. AllOccurrencesFinder.traverseForMatches - reuses the syntax tree already fetched in findAllOccurrences instead of re-fetching it from the manager on every matching node.

These eliminate the redundant 2× full-template scan (and the redundant per-match tree lookups) that caused the freezes, while keeping the offered actions identical.

### DIFF
--- a/tst/unit/services/extractToParameter/ExtractToParameterProvider.test.ts
+++ b/tst/unit/services/extractToParameter/ExtractToParameterProvider.test.ts
@@ -163,6 +163,13 @@ describe('ExtractToParameterProvider', () => {
             const result = provider.canExtract(mockContext);
             expect(result).toBe(false);
         });
+
+        it('should return true for extractable literal in the Outputs section', () => {
+            (mockContext as any).section = TopLevelSection.Outputs;
+
+            const result = provider.canExtract(mockContext);
+            expect(result).toBe(true);
+        });
     });
 
     describe('generateExtraction method', () => {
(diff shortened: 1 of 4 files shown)


tst/unit/services/extractToParameter/ExtractToParameterProvider.test.ts
```diff
@@ -580,8 +587,8 @@ Resources:
         });
     });
 
-    describe('hasMultipleOccurrences method', () => {
```
satyakigh, line 583: Not sure why this diff looks so weird, but it should just be deletion of `'hasMultipleOccurrences method'` and adding new tests

atennak1 (approved):

kddejong (commented):
Disregard.

kddejong (approved):

PR #601, opened 2026-06-10
Sign releases

satyakigh:
* Updating action versions
* Adding signature to release process

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,7 @@ jobs:
       tag: ${{ steps.validate-tag.outputs.tag }}
       is-prerelease: ${{ steps.validate-tag.outputs.is-prerelease }}
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           ref: ${{ inputs.tag }}
           fetch-tags: true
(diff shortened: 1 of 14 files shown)


.github/workflows/release.yml
```diff
@@ -304,12 +306,72 @@ jobs:
...
+          echo "public key SHA-256: $fingerprint"
+
       - name: Create GitHub Release
-        uses: softprops/action-gh-release@v2
+        uses: softprops/action-gh-release@v3
         with:
           tag_name: ${{ needs.version-and-tag.outputs.tag }}
           prerelease: ${{ needs.version-and-tag.outputs.is-prerelease }}
-          files: artifacts/*.zip
+          files: artifacts/*
```
Zee2413, line 374: Nothing unwanted from this change make it in?

.github/workflows/release.yml
```diff
@@ -304,12 +306,72 @@ jobs:
...
+          SIGNING_PUBLIC_KEY: ${{ secrets.SIGNING_PUBLIC_KEY }}
+        run: |
+          set -euo pipefail
+          shopt -s nullglob
+          printf '%s\n' "$SIGNING_PUBLIC_KEY" > signing-key.pem
+
+          for f in artifacts/*; do
+            case "$f" in *.sig) continue ;; esac
+            echo "Verifying: $f"
+            result=$(openssl dgst -sha256 -verify signing-key.pem -signature "$f.sig" "$f") || true
```
Zee2413, line 350: why `|| true`

Zee2413 (approved):

.github/workflows/release.yml
satyakigh: ya everything here needs to be published

.github/workflows/release.yml
satyakigh: probably not required, but its to make the line not fail right away if openssl exists

PR #619, opened 2026-07-02
Add missing metrics and remove capture from private schemas task

satyakigh:
* Remove `captureErrorAttributes` from `GetPrivateSchemasTask`
* Add missing metrics
* Fixing settings sync for CfnService
* Remove unused service


src/services/S3Service.ts
```diff
@@ -104,6 +108,7 @@ export class S3Service {
                 // (network, credentials, throttling, 5xx) propagate unchanged.
                 const { httpStatus } = classifyAwsError(error);
                 if (httpStatus === 403 || httpStatus === 404) {
+                    this.telemetry.error('verifyBucketAccessibleInRegion.error', error);
```
Zee2413: This metrics is a little misleading as an error since this condition will be hit correctly when bucket is rightly inaccessible.

We can emit above new line 117 if we think its useful metric

Zee2413 (dismissed):

Zee2413 (dismissed):

chrisqm-dev (dismissed):

chrisqm-dev (approved):

Issue #624, opened 2026-07-07
[Bug]: Nested Stacks have TemplateURL set to s3:// URI instead of https:// URI


mbarneyjr:
### Describe the bug

It seems the artifact exporter / "packager" functionality seems to "package" `AWS::CloudFormation::Stack` resources incorrectly. The produced template should have https:// URIs defined at `TemplateURL`, but this language server produces S3:// URIs, which produce the following CloudFormation error:

```
S3 error: Domain name specified in <bucket name> is not a valid S3 domain
```

### Steps to reproduce

Define two CloudFormation templates with:
```yml
# template1.local.yml
AWSTemplateFormatVersion: '2010-09-09'
Resources:
  Stack:
    Type: AWS::CloudFormation::Stack
    Properties:
      TemplateURL: ./template2.local.yml
# template2.local.yml
AWSTemplateFormatVersion: '2010-09-09'
Resources:
  Bar:
    Type: AWS::S3::Bucket
```

Call the `aws/cfn/stack/validation/create` method with the following:

```json
{
  "id": "cfn-lsp-test",
  "stackName": "cfn-lsp-test",
  "uri": "/path/to/template1.local.yml",
  "keepChangeSet": true,
  "s3Bucket": "<bucket-name>"
}
```

Then inspect the created ChangeSet's template, you'll see this:

```yml
Stack:
  Type: AWS::CloudFormation::Stack
  Properties:
    TemplateURL: >-
      s3://<bucket-name>/artifact/template2.local-1783391369141.yml

```

### Expected behavior

I would expect to see this be in the produced ChangeSet template:

```yml
Stack:
  Type: AWS::CloudFormation::Stack
  Properties:
    TemplateURL: >-
      https://s3.${region}.amazonaws.com/${bucketName}/...
```

### CloudFormation template (if applicable)

```yaml

```

### Language Server version

1.7.0

### Editor/IDE

Neovim

### Operating System

macOS

### Node.js version

24.16.0

### Logs

```text

```

satyakigh:
Thanks for the finding this bug, the issue should be fixed in https://github.com/aws-cloudformation/cloudformation-languageserver/pull/625 and will be published in the next release

mbarneyjr:
Thank you!

PR #625, opened 2026-07-08
Fix deployment bug for nested templates

satyakigh:
When you have a nested stack - a CloudFormation template that references another template file - the language server has to upload that child template to S3 before it can deploy. It then rewrites the TemplateURL in your template to point at the uploaded copy.

**Problem**: it wrote that URL in the wrong format. It used an s3:// address: `TemplateURL: s3://my-bucket/artifact/template2.yml`. So deployment failed with: `Domain name specified in <bucket name> is not a valid S3 domain`. Nested stacks were effectively broken when uploading through the language server.

**Fix**:  Change the nested-stack TemplateURL to use an https:// address instead: `TemplateURL: https://s3.us-east-1.amazonaws.com/my-bucket/artifact/template2.yml`. Matches what the official AWS CLI does.

- Only nested-template locations were changed (AWS::CloudFormation::Stack TemplateURL and the SAM AWS::Serverless::Application Location). Other uploads - like Lambda code, correctly keep using s3://, because those services do accept it.
- The URL is now region-aware, which also fixed a smaller problem where the top-level template URL was hardcoded to a single global endpoint.

https://github.com/aws-cloudformation/cloudformation-languageserver/issues/624

### DIFF
--- a/src/artifactexporter/ResourceExporters.ts
+++ b/src/artifactexporter/ResourceExporters.ts
@@ -7,6 +7,7 @@ import { dump } from 'js-yaml';
 import { detectDocumentType } from '../document/DocumentUtils';
 import { S3Service } from '../services/S3Service';
 import { readFileIfExists } from '../utils/File';
+import { toHttpsPathStyleS3Url } from '../utils/S3Url';
 import { ArtifactExporter } from './ArtifactExporter';
 
 export function isS3Url(url: string): boolean {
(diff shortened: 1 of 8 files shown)


Zee2413 (dismissed):

src/artifactexporter/ResourceExporters.ts
```diff
@@ -265,9 +266,10 @@ class CloudFormationStackResource extends Resource {
 
         const key = getS3Key(s3KeyPrefix, templateAbsPath);
         await this.s3Service.putObjectContent(exportedTemplateStr, bucketName, key);
-        const s3Url = `s3://${bucketName}/${key}`;
 
-        resourcePropertyDict[this.propertyName] = s3Url;
+        // CloudFormation requires an HTTPS URL for nested-template locations (TemplateURL / Location);
+        // an s3:// URI is rejected with "Domain name specified in <bucket> is not a valid S3 domain".
```
Zee2413: Does this work for the non-nested use cases for artifact upload/large template upload as well?

Zee2413 (dismissed):

src/artifactexporter/ResourceExporters.ts
satyakigh: This class is only used when its a Stack so it this targets nested templates only https://github.com/aws-cloudformation/cloudformation-languageserver/blob/main/src/artifactexporter/ResourceExporters.ts#L390

Zee2413 (approved):

atennak1 (approved):

chrisqm-dev (approved):

PR #631, opened 2026-07-09
Update workflows so ff and manifest checks always run

satyakigh:
* Run generate PRs after long running tests
* Run feature flag and manifest file checks as part of PR workflow so that they are gated by GitHub merge settings

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -79,3 +79,57 @@ jobs:
       - name: Test
         shell: bash
         run: GOPROXY=direct go test -C ./cfn-init -v -cover ./...
+
+  detect-changes:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      release-manifest: ${{ steps.filter.outputs.release-manifest }}
+      feature-flags: ${{ steps.filter.outputs.feature-flags }}
+    steps:
+      - name: Detect changed paths
+        id: filter
+        uses: dorny/paths-filter@v4
+        with:
+          filters: |
+            release-manifest:
+              - 'assets/release-manifest.json'
+            feature-flags:
+              - 'assets/featureFlag/*.json'
+
+  validate-release-manifest:
+    needs: [ detect-changes ]
+    if: needs.detect-changes.outputs.release-manifest == 'true'
+    uses: ./.github/workflows/validate-release-manifest.yml
+
+  validate-feature-flags:
+    needs: [ detect-changes ]
+    if: needs.detect-changes.outputs.feature-flags == 'true'
+    uses: ./.github/workflows/validate-feature-flags.yml
+
+  # Single always-running gate so branch protection can require one stable
+  # check. The validate-* jobs are conditional and report "skipped" on PRs that
+  # don't touch their paths; a skipped job used directly as a required check
+  # would leave the PR pending forever. This job runs unconditionally and only
+  # fails if an upstream job actually failed or was cancelled (skipped is OK).
+  # Make "PR Checks" the single required status check in branch protection.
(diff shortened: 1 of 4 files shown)


Zee2413 (approved):

.github/workflows/pr.yml
```diff
@@ -79,3 +79,56 @@ jobs:
...
+    needs: [ detect-changes ]
+    if: needs.detect-changes.outputs.feature-flags == 'true'
+    uses: ./.github/workflows/validate-feature-flags.yml
+
+  # Single always-running gate so branch protection can require one stable
+  # check. The validate-* jobs are conditional and report "skipped" on PRs that
+  # don't touch their paths; a skipped job used directly as a required check
+  # would leave the PR pending forever. This job runs unconditionally and only
+  # fails if an upstream job actually failed or was cancelled (skipped is OK).
+  # Make "PR Checks" the single required status check in branch protection.
```
kddejong, line 117: **Blocking concern:** This comment tells reviewers to make `pr-checks` the *single* required status check, but `pr-checks` only `needs` `detect-changes` + the two validators — not `pr-build-test-nodejs` / `pr-build-test-go`.

If someone follows this literally and sets `pr-checks` as the only required check, a PR that breaks the TypeScript build or fails Go tests will still show a green `pr-checks` and become mergeable — build/test would silently stop gating merges.

Fix — either:
- add `pr-build-test-nodejs` and `pr-build-test-go` to this job's `needs`, so the gate truly reflects all required checks, **or**
- reword to "add `pr-checks` as an *additional* required check alongside `pr-build-test-nodejs` / `pr-build-test-go`."

Also worth stating the exact check name to configure (the job id is `pr-checks`, not "PR Checks") to avoid a never-satisfied → forever-pending PR.

.github/workflows/release.yml
```diff
@@ -366,15 +366,15 @@ jobs:
...
-      pull-requests: write
-
   long-running-tests:
     needs: [version-and-tag, release]
     uses: ./.github/workflows/post-release-long-running.yml
     with:
       tag: ${{ needs.version-and-tag.outputs.tag }}
+
+  update-release-manifest:
+    needs: [version-and-tag, release, long-running-tests]
```
kddejong, line 376: Adding `long-running-tests` to `needs` means the manifest-update PR is now created only after the 3-OS × 3-node stability matrix finishes (`duration=4h` / `timeout=360` in post-release-long-running.yml). Two things to confirm are intentional:

1. **Delay:** `assets/release-manifest.json` won't reflect the new version for up to ~4h post-release. If any consumer reads the manifest to discover the latest release, that's a behavior change (PR description suggests this is the goal, so likely fine).
2. **Gating:** if any stability leg fails, `long-running-tests` fails and this job is *skipped* — so no manifest update on a failed stability run. Confirm "don't update the manifest if stability tests fail" is desired vs. updating regardless.

kddejong (commented):
Reviewed the workflow changes. The always-running aggregator pattern is sound and the inline explanation is clear. Left two inline comments — one blocking concern on the branch-protection instruction, and one to confirm the intentional delay/gating on the manifest update. Everything else (detect-changes permissions, the pr-checks result loop, removing run-name from the reusable validators, no orphaned triggers) looks correct.

.github/workflows/pr.yml
satyakigh: the main branch still requires the nodejs and go tests to pass, its those + pr-checks for PR to be mergeable

.github/workflows/release.yml
satyakigh: Yup we don't want to create the manifest PR if the tests fail. This is intentional so that the PR can never be merged before all tests are green

PR #635, opened 2026-07-13
Handle compromised advisory file lock instead of crashing

gemammercado:
## Problem

When `proper-lockfile` reports a compromised lock, the process crashes with an uncaught exception.

## Root cause

`LOCK_OPTIONS` in `src/utils/LocalFile.ts` never set `onCompromised`, so `proper-lockfile` uses its default handler, which rethrows (`default: (err) => throw err`). That rethrow fires from `proper-lockfile`'s internal mtime-refresh `setInterval` — outside any awaited call stack — so it surfaces as an `uncaughtException` that no surrounding `try/catch` can intercept.

It triggers when a held lock exceeds the 10s `stale` window (host sleep/suspend) or is stolen by a concurrent process, which also produces the `ENOENT`/`EINVAL utime/stat` variants when the stolen `.lock` is removed out from under the refresh.

`src/datastore/lmdb/OwnershipTracker.ts` imports the same `LOCK_OPTIONS`, so it shares the same gap (rarely hit, it locks once per startup).

## Fix

Set `onCompromised` on the single shared `LOCK_OPTIONS` to log the error with context and continue, instead of rethrowing. This is safe because the write path is already atomic (temp -> `fsync` -> `rename`) and reads are lockless, so a compromised advisory lock cannot corrupt state. Because both consumers share `LOCK_OPTIONS`, this one change covers both `LocalFile` and `LMDBOwnershipTracker`.

## Testing

- `npm run build`, `npm run lint`, and `npm run test` pass.
- Added unit tests in `tst/unit/utils/LocalFile.test.ts` asserting `onCompromised` is defined, does not rethrow, and logs the original error.

## Client-side changes

None required.


src/utils/LocalFile.ts
```diff
@@ -30,6 +30,12 @@ export const LOCK_OPTIONS: LockOptions = {
         maxTimeout: 2000, // Maximum delay (ms) cap for any single retry
         randomize: true, // Add jitter to retry delays to avoid contention between concurrent writes
     },
+    onCompromised: (error) => {
```
satyakigh, line 34: Will this continue to fail? Do we need a recovery mechanism? Also emit a metric

gemammercado: No, it self-recovers (`proper-lockfile` deletes the in-memory lock so the next write re-acquires a fresh one, and the stale `.lock` is reclaimed via the `stale` value (10 seconds) in `LOCK_OPTIONS` stale-steal path). No recovery mechanism is needed since writes are atomic and reads are lockless. The only residual is that the write coinciding with the compromise rejects with ERELEASED, which the caller already catches and logs. I'll also add a lock.compromised metric

satyakigh (approved):

chrisqm-dev (approved):

PR #638, opened 2026-07-14
Add generic error categories and add more aws classifications

satyakigh:
* Fix histogram data for document lines
* Add more aws error categories
* Add generic error categories


src/utils/errors/ErrorStackInfo.ts
```diff
@@ -47,14 +48,20 @@ export function errorType(error: unknown): Attributes {
...
     }
-    if (awsClassification.httpStatus) {
+    if (awsClassification.httpStatus !== undefined) {
         awsAttr['error.aws.http.status'] = sanitizeMessage(`${awsClassification.httpStatus}`);
     }
 
+    const genericCategory = classifyGenericError(error);
+    const genericAttr: Record<string, string> = {};
+    if (genericCategory !== undefined) {
+        genericAttr['error.category'] = sanitizeMessage(genericCategory);
```
chrisqm-dev, line 58: Should we add the sanitizeMessage at line 64 instead of here to be consistent with the other attributes? Same with error.aws.category

chrisqm-dev (approved):

src/utils/errors/ErrorStackInfo.ts
satyakigh: Think its fine

PR #641, opened 2026-07-16
Expose client id in lsp client

satyakigh:
* Expose client id for usage in tests

### DIFF
--- a/tools/lspClient/LspClient.ts
+++ b/tools/lspClient/LspClient.ts
@@ -12,10 +12,10 @@ import {
     MessageConnection,
     TextDocumentContentChangeEvent,
 } from 'vscode-languageserver-protocol';
-import { randomBytes } from 'crypto';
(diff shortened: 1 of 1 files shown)


tools/lspClient/LspClient.ts
```diff
@@ -215,7 +222,7 @@ export class LspClient implements LspConnection {
         return this.connection.sendRequest(method, params);
     }
 
-    sendNotification(method: string, params: unknown): Promise<void> {
+    sendNotification(method: string, params: unknown) {
```
gemammercado, line 225: is there any risk with removing the promise annotation?

gemammercado (approved):

tools/lspClient/LspClient.ts
satyakigh: Nope

PR #653, opened 2026-07-28
Refactor and improve datastores, improve metrics

satyakigh:
The persisted datastore layer degraded badly under I/O and recovery failures, and the telemetry did not report those failures well enough to diagnose them.

1.  FileDB updated in-memory state before the disk operation finished, so a full disk or a permission error left a session serving values that were never written
2.  LMDB recovery reopened its environment without releasing the old handle, which consumed finite reader slots, and it deleted database files before the asynchronous close finished
3. The real cause of an LMDB write failure sat behind an unresolved `commitError` promise, so a full disk reached telemetry as a generic `Commit failed` error

#### FileDB served values it never persisted
* **Bug** — The file store updated memory before awaiting the write or unlink, so a failed disk operation left the session serving unwritten values and hiding keys whose files still existed.
* **Fix** — Disk first, memory only on success, across all four methods.

#### LMDB recovery leaked environment handles
* **Bug** — Reopen replaced the environment without closing the old handle, leaking one finite reader slot per recovery until the store stopped working.
* **Fix** — Open the replacement, then release the previous handle. A partial failure releases both.

#### Startup recovery deleted files before the close finished
* **Bug** — Recovery deleted the version directory without awaiting the close, which Windows rejects while the file is open.
* **Fix** — Await the release before deleting. The test asserts close, then delete.

#### A failed startup left a stale ownership marker
* **Bug** — A JavaScript-level open failure left the opening marker behind, so the next launch read it as a prior crash and wiped intact data.
* **Fix** — Release the marker on failed initialization. Real crashes carry a sentinel error, so telemetry reports a prior crash instead of corruption.

#### Disk-full failures reached telemetry as generic errors
* **Bug** — lmdb-js hides the real write error behind a rejected promise, and the extraction tested for an Error instance, so out-of-space never reached telemetry and the rejection went unconsumed. Numeric error codes were ignored too.
* **Fix** — Resolve that promise within 500 ms, promote it to the standard cause, and emit per-operation out-of-space and free-space metrics from both factories.

#### Shutdown produced faults, and timers held the process
* **Bug** — The datastore closed before telemetry, so the final metric flush read a closed database. The memory factory was never closed, and its interval held the event loop open.
* **Fix** — Close telemetry first and the datastore last, close both factories, and unreference the timers.

#### Schema-age gauges reread the whole store
* **Bug** — Every metric export re-read and decrypted full schema payloads for one timestamp, and an empty store reported age zero, which reads as freshly downloaded.
* **Fix** — Cache the timestamp per region, clear it after a download, and report an empty store as the maximum safe integer.

#### LMDB
* Opening the database and binding its stores were two operations, called in different combinations by each recovery path. They are now one, so a cold start and a recovery run the same code. Stores are rebound in place, so a caller holding one keeps working across a recovery.
* Recovery used to throw away the error it caught, and the delete-and-reopen step swallowed its own failure, so two failures in a row left a dead store behind one log line. Both now keep the error, hand it to the code that classifies data loss, and rethrow. Without that cause, every wipe is reported as unknown.
* Fork detection no longer carries recovery inside it. The check handed to each store is a process-id comparison that delegates elsewhere, which also collapsed two metric names for one event into one.

#### Metrics

Both stores now count out-of-space failures per operation as `{operation}.enospc`, count discarded data under twelve reasons as `store.discarded.{reason}`, and sample the filesystem holding their data for `disk.available.bytes`, `disk.total.bytes`, and `disk.available.percent`. A new percent aggregation view keeps a free-space floor from being averaged away.

### Utilities
* A new chain walker returns every link from an error through its causes, capped at eight links and guarded against cycles. Root-cause extraction now reads the second link from that walker instead of hand-checking two properties. Error-code extraction gained numeric `code` values, which is how the native LMDB layer reports out of space as 28. Previously only a numeric `errno` was read.
* `CommitError` - Unwraps the error lmdb-js hides behind a promise and promotes it to the standard cause property, where the chain walker finds it.
* Recognizes out of space from `ENOSPC`, numeric 28 and −28, and three message fragments, and recognizes eight inaccessible-path codes including the Windows lock signal. Both search the whole cause chain. Free space comes from a filesystem stat, counting blocks available to unprivileged writers, and returns nothing on a mount without support so a metrics cycle never faults.
* `Utils` — Sorts a failure into one of twelve named reasons for data loss and holds the three recorders both stores share. Environmental causes outrank corruption: out of space and permission failures are searched across the full chain before any corruption verdict, so a full disk buried inside a corruption report still reports as a full disk.


src/datastore/LMDBStoreFactory.ts
```diff
@@ -193,65 +176,40 @@ export class LMDBStoreFactory implements DataStoreFactory {
...
         }
     }
 
-    private deleteAndRecreate(): void {
+    private deleteAndRecreate(cause: unknown): OpenedEnv {
         try {
-            this.deleteVersionDir();
-            this.reopenEnv();
-            this.recreateStores();
+            this.deleteVersionDir(cause);
```
kddejong, line 236: `releaseEnv` is fire-and-forget in `reopenEnv`, so this `deleteVersionDir` can `rmSync` the data dir while the previous env's `close()` still holds `data.mdb` -> EBUSY/EPERM on Windows. The `releaseEnv` doc says delete-path callers should await the close, but `recoverFromFork`/`recoverFromError` reach here synchronously. Suggest making this async and awaiting the release first:

```ts
private async deleteAndRecreate(cause: unknown): Promise<OpenedEnv> {
    await this.releaseEnv(this.env);
    this.deleteVersionDir(cause);
    return this.createEnvAndStores();
}
```

src/datastore/file/KeyedFileStore.ts
```diff
@@ -146,6 +162,19 @@ export class KeyedFileStore implements DataStore {
...
     }
+
+    private createEncryptedFile(fileName: string) {
+        const file = new LocalFile(join(this.fileDbDir, fileName));
+
+        try {
+            return new EncryptedFile(this.encryptionKey, file);
+        } catch (e) {
+            this.onDiscard(e);
+            file.unsafeRemove();
```
kddejong, line 182: `unsafeRemove()` (`unlinkSync`) can throw EACCES/EBUSY on Windows if a handle is still open -- uncaught here, so one bad file crashes `loadAllFiles`. And if the remove silently no-ops on a still-corrupt file, the retry `new EncryptedFile(...)` re-throws the same decrypt error. Suggest wrapping it:

```ts
try {
    file.unsafeRemove();
} catch (removeErr) {
    this.log.warn(removeErr, `Failed to remove unreadable file ${fileName}`);
}
```

tst/unit/datastore/LMDB.enospc.test.ts
```diff
@@ -0,0 +1,118 @@
...
+        await factory.initialize();
+    });
+
+    afterEach(async () => {
+        vi.restoreAllMocks();
+        await factory.close();
+        fs.rmSync(testDir, { recursive: true, force: true });
+    });
+
+    function failNextWriteWith(error: Error): void {
```
atennak1: nit: slightly misleading. `failAllWritesWith` may be better

src/datastore/FileStoreFactory.ts
```diff
@@ -78,18 +87,27 @@ export class FileStoreFactory implements DataStoreFactory {
...
             });
+
+            totalBytes += stats.totalSize;
         }
 
-        this.telemetry.histogram('total.size.bytes', this.totalBytes(), {
+        this.telemetry.histogram('total.size.bytes', totalBytes, {
             unit: 'By',
         });
+        this.telemetry.histogram('total.usage', 100 * (totalBytes / TotalMaxDatastoreSize), { unit: '%' });
```
atennak1, line 108: should this be `total.usage.percent`?

src/datastore/file/KeyedFileStore.ts
```diff
@@ -17,90 +20,103 @@ export class KeyedFileStore implements DataStore {
...
-            },
-            { captureErrorAttributes: true },
-        );
+        return this.execAsync(StoreOperation.clear, async () => {
+            this.loadAllFiles();
+            const files = [...this.keysToFiles.values()];
+            for (const file of files) {
+                await file.remove();
+            }
+            this.keysToFiles.clear();
```
atennak1: may not get here if one file fails to remove

src/datastore/LMDBStoreFactory.ts
satyakigh: `LMDBStoreFactory` now tracks pending environment releases and awaits them before delete/recreate. Async operations await recovery; concurrent destructive recoveries are deduplicated. A deferred-close test verifies close finished → delete

src/datastore/FileStoreFactory.ts
satyakigh: LMDB uses `total.usage` (pre-existing) so kept this the same. both .usage and .percent will work

src/datastore/file/KeyedFileStore.ts
satyakigh: Fixed, only removing an entry that was successfully deleted

src/datastore/file/KeyedFileStore.ts
satyakigh: `recoverFile()` catches all exceptions so this shouldn't be an issue, and only valid files are tracked in memory - added a test for this usecase

kddejong (approved):

atennak1 (approved):

PR #655, opened 2026-07-30
feat(hooks): add TTL HookCache and persistent hook schema store

gemammercado:
## PR 1 of 7 — foundation (stacked)

This is the first slice of the CloudFormation Hooks feature, split into 7 stacked PRs for reviewability. It adds the low-level building blocks that every later slice depends on. Nothing is wired into the server yet — this PR is pure foundation.
### What this adds

- **`HookCache`.** Memoises hook configurations and Guard rule bodies, so a feature
  that repeatedly inspects the same hook doesn't re-issue a CloudFormation or S3 call
  each time. Configurations and rule bodies live in separate namespaces, each backed
  by [`lru-cache`](https://github.com/isaacs/node-lru-cache) with a TTL and a maximum
  entry count. Concurrent loads of the same key are coalesced into a single call.
- **Hook LSP request types.** The `Params`/`Result`/`Request` definitions for every
  hook-related LSP message. This is the contract the handlers and client use in later
  slices.

### What changed in this revision

Addressing @satyakigh's review. Three of the four comments are resolved by removing
code rather than changing it, so the corresponding inline comments will show as
outdated.

**Persisting hook schemas.** Correct on both counts, and the second was a latent bug.
`CfnService.listHooks()` filters to `Visibility: PRIVATE`, so those records are hook
type names and configuration fields authored inside the customer's own account —
the same category as `private_schemas`, which is already memory-only here. They were
also keyed by type name alone with no account or region qualifier, so a persisted
record could be served after a profile or region switch, and on disk that survived
restarts with a staleness window measured in days.

Rather than move it to memory, `HookSchemaStore` is **removed entirely**. Once
persistence went away, its only distinguishing feature over the schema cache
`HooksManager` already keeps in memory was a staleness window, which an lru-cache
`ttl` expresses directly. The schema cache is therefore better introduced alongside
`HooksManager` than as a separate store here. That also drops the `hook_schemas`
store, so **`DataStore.ts` is no longer touched by this PR at all**.

**Cache size.** Unbounded as written. In practice neither cache was keyed by user activity — configuration keys come from the hooks activated in the account, and rule keys from the rule URIs those configurations reference, so the key count tracked hook count rather than session length. But that was a property of the callers, not of the cache, and rule bodies read from S3 have no inherent size limit. Both namespaces are
now capped (100 entries each by default) with LRU eviction.

**Pruning while iterating in `size`.** The `size` accessor no longer exists. lru-cache owns staleness now, so there is no second traversal to fold in.

**`TtlCache` didn't need exporting.** Correct — nothing outside the module used it.
The hand-rolled `TtlCache` is deleted outright in favour of lru-cache, which supplies
TTL expiry, the entry bound, LRU eviction and request coalescing. `HookCache` is the
only export.

### New dependency

`lru-cache` becomes a direct dependency, pinned exactly to match the other production
dependencies. It was already in the tree transitively via `path-scurry` and is
recorded in `THIRD-PARTY-LICENSES.txt` and `sbom/sbom.csv`.

⚠️ One thing to flag: **v11 is licensed `BlueOak-1.0.0`**, where the transitive
v10.4.3 was ISC — Isaac relicensed at v11. Blue Oak is permissive and OSI-approved,
but it falls outside the set `ATTRIBUTION_README.md` treats as routine
(`MIT, Apache-2.0, BSD-2-Clause, BSD-3-Clause, ISC, 0BSD`), so it will surface in the
license report and wants a conscious sign-off. v10.4.3 is the ISC alternative, but it
lacks the `perf` option this PR uses to inject a clock in tests. SBOM and attribution
still need regenerating.

### Testing

Tests cover what this wrapper contributes rather than re-testing lru-cache: separate
namespaces for configurations and rules, invalidation across both, that the `ttl` and
`maxEntries` options reach each namespace, that each namespace is bounded
independently rather than sharing one budget, that rejected loads stay uncached, and
that concurrent loads of one key are coalesced. Build, lint and unit tests pass.


src/datastore/DataStore.ts
```diff
@@ -17,9 +17,14 @@ export enum StoreName {
...
     sam_schemas = 'sam_schemas',
     private_schemas = 'private_schemas',
+    hook_schemas = 'hook_schemas',
 }
 
-export const PersistedStores: ReadonlyArray<StoreName> = [StoreName.public_schemas, StoreName.sam_schemas];
+export const PersistedStores: ReadonlyArray<StoreName> = [
+    StoreName.public_schemas,
+    StoreName.sam_schemas,
+    StoreName.hook_schemas,
```
satyakigh: Should this be persisted? We don't persist private customer data on disk. Will it change with credentials/regions?

src/hooks/HookCache.ts
```diff
@@ -0,0 +1,132 @@
+interface CacheEntry<T> {
+    value: T;
+    expiresAt: number;
+}
+
+export class TtlCache<T> {
+    private readonly entries = new Map<string, CacheEntry<T>>();
```
satyakigh: How large can these caches get?

src/hooks/HookCache.ts
```diff
@@ -0,0 +1,132 @@
...
+            if (entry.expiresAt <= now) {
+                this.entries.delete(key);
+            }
+        }
+    }
+
+    get size(): number {
+        let count = 0;
+        for (const entry of this.entries.values()) {
+            if (entry.expiresAt > this.now()) {
```
satyakigh: If the whole map is being iterated, could just prune here as well

src/hooks/HookCache.ts
```diff
@@ -0,0 +1,132 @@
+interface CacheEntry<T> {
+    value: T;
+    expiresAt: number;
+}
+
+export class TtlCache<T> {
```
satyakigh: Doesn't need to be exported

kddejong (dismissed):

src/datastore/DataStore.ts
gemammercado: Moved to Persistence.memory and reverted the PersistedStores change.

On credentials/regions: yes, this was a bug. Records were keyed `hook:${typeName}` with no account or region qualifier, so switching profile or region would read back the previous account's schema under the same type name. On disk that survived restarts, and the staleness threshold was in days.

On privacy: `CfnService.listHooks()` filters `Visibility: PRIVATE`, so this store was fed almost entirely by hooks the customer registered into their own account. Same category as private_schemas, which is already memory-only here.

src/hooks/HookCache.ts
gemammercado: Unbounded as written, added a 100-entry cap per cache and oldest-first eviction

src/hooks/HookCache.ts
gemammercado: size now calls prune()

src/hooks/HookCache.ts
gemammercado: un-exported. It was only exported so the unit tests could drive it directly

satyakigh (approved):

kddejong (approved):

PR #662, opened 2026-08-13
feat(hooks): AWS service layer — ControlCatalog, IAM, hook APIs (2/7)

gemammercado:
## PR 2 of 7 — AWS service layer (stacked)

Second slice of the CloudFormation Hooks feature (follows #655). This adds the AWS service-layer pieces the hooks feature calls into. Still no LSP wiring — handlers come in a later slice.

### What's in it

- **`ControlCatalogService`.** Thin wrapper over the AWS Control Catalog API used to list proactive controls when browsing hooks backed by managed rules. New dependency: `@aws-sdk/client-controlcatalog`, pinned `3.1069.0` — this is the latest published version of that client (the other SDK clients are on `3.1080.0`; controlcatalog releases lag the rest of the SDK).
- **`IamService`.** Lists IAM roles and creates the Guard Hook execution role (trusts `hooks.cloudformation.amazonaws.com` scoped to the caller's account; optional inline S3 read policy scoped to the rule bucket, with bucket-name validation before it's embedded in a policy ARN).
- **Shared `AwsClient.getCallerAccountId()`.** STS caller-account resolution extracted to `AwsClient` and used by both `IamService` and `S3Service`.
- **Hook APIs on the existing AWS clients.** `CfnService`: list/describe/activate/deactivate hooks, hook type-configuration get/set, hook results, change-set hooks (paginated). `CcapiService.createResource` (polls CloudControl to a terminal state). `S3Service`: bucket listing/creation and object listing/read for Guard rule files.

### Telemetry
Decorators follow the repo convention: bare `@Count({ name })` / `@Measure({ name })` — no `captureErrorAttributes`, no explicit `captureErrorType`.

### Testing
Unit tests for ControlCatalogService, IamService, and all CfnService hook APIs (success, error, and pagination paths). Build and lint pass.

### Stack
Base: `main` (requires #655, merged). Next slice: HooksManager + HooksParser.

### DIFF
--- a/src/services/AwsClient.ts
+++ b/src/services/AwsClient.ts
@@ -1,5 +1,7 @@
 import { CloudControlClient } from '@aws-sdk/client-cloudcontrol';
 import { CloudFormationClient } from '@aws-sdk/client-cloudformation';
+import { ControlCatalogClient } from '@aws-sdk/client-controlcatalog';
+import { IAMClient } from '@aws-sdk/client-iam';
 import { S3Client } from '@aws-sdk/client-s3';
 import { STSClient } from '@aws-sdk/client-sts';
 import { AwsCredentials } from '../auth/AwsCredentials';
@@ -35,6 +37,14 @@ export class AwsClient {
 
     public getStsClient() {
         return new STSClient(this.iamClientConfig());
+    }
+
+    public getIamClient() {
+        return new IAMClient(this.iamClientConfig());
+    }
+
+    public getControlCatalogClient() {
+        return new ControlCatalogClient(this.iamClientConfig());
     }
 
     public getRegion(): string {
--- a/src/services/CallerIdentity.ts
+++ b/src/services/CallerIdentity.ts
@@ -1,0 +1,21 @@
+import { GetCallerIdentityCommand } from '@aws-sdk/client-sts';
+import { LoggerFactory } from '../telemetry/LoggerFactory';
+import { markIfClientError } from '../utils/errors/FaultSuppression';
+import { AwsClient } from './AwsClient';
+
+const log = LoggerFactory.getLogger('CallerIdentity');
+
+export async function getCallerAccountId(awsClient: AwsClient): Promise<string> {
+    const sts = awsClient.getStsClient();
+    try {
+        const identity = await sts.send(new GetCallerIdentityCommand({}));
+        if (!identity.Account) {
+            throw new Error('STS GetCallerIdentity did not return an account ID');
+        }
+        return identity.Account;
+    } catch (error) {
+        log.error(error, 'Failed to resolve caller account ID via STS');
+        markIfClientError(error);
+        throw error;
+    }
+}
--- a/src/services/CcapiService.ts
+++ b/src/services/CcapiService.ts
@@ -53,7 +53,7 @@ export class CcapiService {
         });
     }
 
-    @Measure({ name: 'getResource' })
+    @Measure({ name: 'getResource', captureErrorType: true })
     public async getResource(typeName: string, identifier: string) {
         return await this.withClient(async (client) => {
             const getResourceInput: GetResourceInput = {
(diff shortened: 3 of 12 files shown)


src/services/CfnService.ts
```diff
@@ -463,6 +503,144 @@ export class CfnService implements SettingsConfigurable, Closeable {
         this.settingsSubscription?.unsubscribe();
         this.settingsSubscription = undefined;
     }
+
+    @Count({ name: 'listHooks', captureErrorAttributes: true })
+    public async listHooks(nextToken?: string): Promise<{ hooks: TypeSummary[]; nextToken?: string }> {
```
satyakigh, line 508: * `captureErrorAttributes: true` can't be used for any online components
* `captureErrorType: true` is enabled by default, doesn't have to be set explicitly

satyakigh (changes requested):

src/services/AwsClient.ts
```diff
@@ -37,10 +43,33 @@ export class AwsClient {
...
+
+    public getControlCatalogClient() {
+        return new ControlCatalogClient(this.iamClientConfig());
+    }
+
     public getRegion(): string {
         return this.credentialsProvider.getIAM().region;
     }
 
+    public async getCallerAccountId(): Promise<string> {
```
satyakigh: Should this be in `IAMService`? This class is just providing aws clients

satyakigh (dismissed):

src/services/AwsClient.ts
gemammercado: agreed AwsClient is a client factory, so it doesn't belong there. I only pulled it up because IamService needed the same logic S3Service already had, and I wanted to avoid duplicating it.

Reverting the AwsClient change: IamService now owns its own private getCallerAccountId(), and S3Service is back to what's on main. That does leave the two services with the same ~12-line STS lookup so I moved it to a helper in `src/services/CallerIdentity.ts`

src/services/CallerIdentity.ts
```diff
@@ -0,0 +1,21 @@
+import { GetCallerIdentityCommand } from '@aws-sdk/client-sts';
+import { LoggerFactory } from '../telemetry/LoggerFactory';
+import { markIfClientError } from '../utils/errors/FaultSuppression';
+import { AwsClient } from './AwsClient';
+
+const log = LoggerFactory.getLogger('CallerIdentity');
+
+export async function getCallerAccountId(awsClient: AwsClient): Promise<string> {
```
chrisqm-dev, line 8: Are we not able to get this from the existing config? We should send over the account id when sending credentials

src/services/CcapiService.ts
```diff
@@ -50,7 +53,7 @@ export class CcapiService {
         });
     }
 
-    @Measure({ name: 'getResource', captureErrorType: true })
```
chrisqm-dev, line 53: why'd we remove captureErrorType from here? Was this cleanup?

chrisqm-dev (dismissed):

src/services/CcapiService.ts
gemammercado: unintentional. getResource is pre-existing and unrelated to hooks. I removed it too broadly when applying the feedback about not setting captureErrorType explicitly on new hook methods. It's a no-op either way (errorType is always captured) but it doesn't belong in this PR, so I'm reverting it

satyakigh: This is enabled by default now, safe to remove

src/services/CallerIdentity.ts
gemammercado: DecryptedCredentialsSchema doesn't include accountId, so zod strips it even if the client sends it today. I could couple this with a client side change in a follow up PR as `S3Service.verifyBucketAccessibleInRegion` also does an STS roundtrip.
We'd need to keep the STS lookup as a fall back since older clients wouldn't send accountId

chrisqm-dev (approved):

satyakigh (approved):

PR #676, opened 2026-09-04
Add Metadata.Context completion, hover, and quick fix

satyakigh:
## Summary

Add first-class editor support for CloudFormation `com.aws.cloudformation.Context` metadata.

* Add schema-aware completion for:
 * Template fields: `arch`, `must`, `ref`, and `owner`
 * Resource fields: `why`, `must`, `mutable`, `mutability`, `trust`, and `deps`
 * Nested `ref` and `trust` structures
 * Mutability, trust-source, and confidence enum values
* Add hover documentation with field types, required fields, enum values, advisory-schema guidance, and links to the authoritative AWS documentation.
* Add a quick fix that inserts template- or resource-level Context skeletons in response to missing-context diagnostics.
* Support YAML and JSON while preserving existing arbitrary Metadata entries.
* Support recoverable malformed templates, including partial keys, missing JSON closing braces, and valid islands following unrelated syntax errors.
* Preserve existing behavior at wholly invalid positions and avoid offering CodeActions when a safe insertion point cannot be resolved.

The shared Metadata Context schema uses the existing `PropertyType` convention under `src/schema`. No LSP protocol or client changes are required.

Documentation: https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-metadata.html#aws-attribute-metadata-context-schema

## Implementation notes

Malformed JSON path recovery now:

* Runs only when the cursor has a local tree-sitter `ERROR` ancestor
* Preserves normal ancestor and entity context
* Merges normal and fallback property paths only when they overlap
* Distinguishes partial object keys from scalar values
* Handles UTF-8 tree-sitter columns without copying the full document


src/context/syntaxtree/SyntaxTree.ts
```diff
@@ -678,25 +724,24 @@ export abstract class SyntaxTree {
...
+            }
+            currentNode = currentNode.parent;
         }
 
         if (!errorNode) {
             return { path, propertyPath, entityRootNode };
         }
 
-        const text = errorNode.text;
+        const text = this.tree.rootNode.text;
```
satyakigh: [P2] Handle even-length backslash runs while scanning JSON strings

This now scans the whole document prefix, so it encounters earlier valid strings. The quote-terminator check below only examines the immediately preceding character; therefore it treats the closing quote in a valid JSON value ending in an escaped backslash (`"Description": "C:\\"`) as escaped, even though the two-backslash run means the quote terminates the string. With missing closing braces at a later Context key, my probe returned `["Description", "ar"]` instead of `["Metadata", "com.aws.cloudformation.Context", "ar"]`, so the new malformed-template completion/hover support disappears. Please count the full preceding backslash run (odd means escaped), or reuse a JSON lexer, and add this case to `JsonFallback.test.ts`.

src/services/MetadataContextQuickFix.ts
```diff
@@ -0,0 +1,325 @@
...
+        private readonly documentManager: DocumentManager,
+        private readonly contextManager: ContextManager,
+    ) {}
+
+    static isMissingContextDiagnostic(diagnostic: Diagnostic): boolean {
+        const normalizedCode = String(diagnostic.code ?? '')
+            .trim()
+            .toUpperCase()
+            .replaceAll(/[-.\s]+/g, '_');
+        if (missingContextDiagnosticCodes.has(normalizedCode)) {
```
satyakigh: [P2] Gate generic diagnostic codes by Metadata-specific evidence

`MISSING_CONTEXT` is accepted solely by code, regardless of `source` or message. Guard diagnostic codes are user-defined rule names, so a `cfn-guard` rule named `MISSING_CONTEXT` with message `Required application context is missing` now receives an unrelated **Add Metadata.Context** action; the executable probe confirmed that it inserts this block into the resource. Please require `MISSING_METADATA_CONTEXT_DIAGNOSTIC_SOURCE` for generic aliases, or require the message to name Metadata Context, while retaining the explicit cfn-lint full-key message match.

src/services/MetadataContextQuickFix.ts
```diff
@@ -0,0 +1,325 @@
...
+
+        const context = this.contextManager.getContext({
+            textDocument: { uri },
+            position: diagnostic.range.start,
+        });
+        if (context?.section === TopLevelSection.Resources && context.logicalId) {
+            return { scope: 'resource', logicalId: context.logicalId };
+        }
+
+        return { scope: 'template' };
```
satyakigh: [P1] Do not change an unresolved resource diagnostic into a template edit

For diagnostic data `{ scope: 'resource' }` without a `logicalId`, when the range is on `Resources` and therefore has no contextual logical ID, execution reaches this default. The resulting action inserts template-level `Metadata` with `arch`, despite the caller explicitly declaring resource scope. This contradicts the stated safety behavior of declining edits whose target cannot be resolved. Return `undefined` for an unresolved explicit resource scope (or resolve its logical ID) rather than silently switching scopes.

satyakigh (commented):
.

src/services/MetadataContextQuickFix.ts
```diff
@@ -0,0 +1,325 @@
...
+
+        const valueNode = NodeType.extractValueFromPair(pairNode, documentType);
+        if (!valueNode) {
+            return undefined;
+        }
+        if (NodeType.isMappingNode(valueNode, documentType)) {
+            return valueNode;
+        }
+
+        return valueNode.descendantsOfType(
```
satyakigh: [P1] Restrict insertion to the pair's actual mapping value

Searching all descendants lets non-mapping values masquerade as the target mapping. For a sequence-valued `Metadata`, the action inserts beside/inside a nested item; for an `Fn::ForEach` resource, the direct resource path resolves to the intrinsic sequence and this inserts `Metadata` between sequence items. Both focused probes produced YAML parse errors. When there is no descendant, `insertIntoPairValue` also treats `Metadata: custom-value` like an empty value and produces invalid YAML. Please only unwrap the YAML structural wrapper to a direct mapping, reject non-empty scalar/sequence values, and either resolve the nested ForEach resource path explicitly or offer no edit.

src/services/MetadataContextQuickFix.ts
```diff
@@ -0,0 +1,325 @@
...
+        if (mappingNode) {
+            return this.insertIntoMapping(mappingNode, propertyName, propertyValue, document);
+        }
+
+        if (document.documentType !== DocumentType.YAML) {
+            return undefined;
+        }
+
+        const indentation = ' '.repeat(pairNode.startPosition.column + document.getTabSize());
+        const newText = `\n${this.formatYamlProperty(propertyName, propertyValue, indentation, document.getTabSize())}`;
```
satyakigh: [P2] Preserve an inline comment on an empty Metadata key

For `Metadata: # preserve metadata explanation`, the pair ends before the comment and this inserted text has no trailing newline. Applying the action moves the existing comment onto `mutable: review-required`, changing what the comment documents. The probe reproduced exactly that output. Please insert after the original commented line (or otherwise account for the trailing comment) so the generated block cannot absorb it, and cover this case in the quick-fix tests.

Zee2413 (approved):

kddejong (approved):

PR #679, opened 2026-09-09
feat(hooks): local Guard hook preview (4/7)

gemammercado:
## PR 4 of 7 — local Guard hook preview (stacked)

Fourth slice of the CloudFormation Hooks feature, following #655, #662, and #674 (all merged). It answers the question "what would my Guard Hooks say about this template?" locally, without deploying anything or creating a change set. LSP handlers and wiring come in the next slice, so nothing here is reachable from the server yet.

### Scope: stack-target hooks only

The preview evaluates only hooks that would run at the stack level — `targetOperations` containing `STACK`, or left empty/undefined. Hooks targeting only `RESOURCE`, `CLOUD_CONTROL`, or `CHANGE_SET` are deliberately excluded rather than evaluated against the wrong input, because each target receives a differently shaped input document (per-resource properties, `Previous`, `Changes`) and feeding a raw template to those rules produces false passes and false failures.

This also keeps the preview faithful rather than approximate. A stack invocation receives the template *as passed to* `create-stack`/`update-stack`, so its intrinsics (`!Sub`, `!Ref`) are unresolved — exactly what we read off disk. Resource invocations instead receive `ResourceProperties` with functions, parameters, and transforms fully resolved, which a local preview cannot reproduce for `!GetAtt` or `!ImportValue`. Supporting those targets requires deciding how to handle that gap, plus which `InvocationPoint` to assume when no operation is in flight and where `Previous` would come from; that is written up as follow-up work rather than guessed at here.

### What's in it

**`GuardHookPreview`** — takes the configured hooks from `HooksManager`, keeps the ones that would actually run (configured, `ENABLED`, with a rule location, stack-targeted), fetches each hook's Guard rule, and evaluates it against the open template using the same engine the editor already uses for Guard diagnostics. Each entry reports pass/fail plus the individual violations with their locations, so the result mirrors what the deployed hook would report.

- A hook's rule is checked for validity before evaluation. This matters because cfn-guard cannot report it for us: its WASM binding has a single entry point (`tryBuildAndExecute`) that returns an empty result set for malformed, empty, or nonsense rules — indistinguishable from a rule that genuinely passed. Without the check, a broken rule would be reported to the user as "this hook passes". A rule that fails the check is reported as invalid with the reason instead.
- Severity follows the hook's failure mode, so a `WARN` hook previews as warnings and a `FAIL` hook as errors.
- One failing hook (an unreadable rule file, say) is reported against that hook only and does not fail the whole preview. The cause is logged server-side rather than only returned inline.
- Rule fetches are batched at a concurrency of 10, and hooks that share a rule location fetch it once.

**`GuardHookInput`** — builds the input document the rules are evaluated against, attaching the `HookContext` block that stack-target rules can query. It models a create: `InvocationPoint` is `CREATE_PRE_PROVISION` and `Previous` is absent, which is the documented shape for a create. Fields with no local equivalent (`StackId`, `ChangeSetId`, `TargetName`) are empty strings rather than invented values.

Field casing was settled against the Guard hook application's own test fixture, which uses `AwsAccountId`, because the public documentation contradicts itself — its syntax tables say `AWSAccountID` while its examples say `AwsAccountId`.

YAML templates are appended to textually instead of being parsed and re-serialized. Re-serializing would rewrite short-form intrinsics (`!Ref`, `!Sub`) into their long form or mangle them outright, changing the very content the rules inspect. JSON templates are parsed and merged; if that parse fails — a template mid-edit, or YAML that happens to open with `{` — the original text is returned unchanged and the reason is logged, so the fallback is visible rather than silent.

**`GuardEngine.evaluateRule`** — a single entry point that runs the syntactic pre-check and the evaluation once and returns `{ valid, parseErrors, violations }`. `validateRule` is now a thin wrapper over it (defaulting the data to an empty template) and the preview delegates to it, so a rule is never validated and then re-evaluated. The return type is exported so the two call sites cannot drift.

The pre-check itself is best-effort, because cfn-guard surfaces no parse or compile diagnostics. It catches the common authoring mistakes — empty or comments-only content, unbalanced or mismatched brackets — while deliberately ignoring bracket and quote characters inside strings, regex literals (including regex character classes, where a `/` is not a terminator), `#` comments, and `<< ... >>` custom-message blocks, so a legitimate rule whose message contains punctuation is not rejected. It cannot detect every invalid rule: a bracket-balanced but meaningless rule still evaluates to zero violations. It is a guard against obvious mistakes, not a substitute for a real parser.

### Testing

Unit tests cover the preview end to end against a real `GuardEngine` and a multi-resource template, the invalid-rule path, per-hook error isolation, severity selection, shared-rule deduplication, target-operation filtering, and the empty case.

`evaluateRule`/`validateRule` tests cover the pre-check's accept and reject cases: messages containing brackets and apostrophes, unterminated quotes, strings, regex literals including a character-class case read from the vendored `advanced_regex_negative_lookbehind_rule.guard` fixture, unterminated regexes, an unclosed `<< ... >>` block, a string line ending in a backslash, and comments — plus evaluation with and without sample data.

`GuardHookInput` tests cover both the JSON and YAML paths, including an empty template, a template with no `Resources`, CRLF line endings, and a template that already carries a `HookContext` — where the JSON path overwrites and the YAML path appends, pinned by tests so the divergence is deliberate rather than accidental.

Verified manually end to end as well, against a real `AWS::Hooks::GuardHook` with its rule in S3: both violations were reported against the correct resources with their custom messages intact, and the compliant resource stayed clean.

Build, lint, and tests pass.

### Stack

Base: `main` (requires #655, #662, #674, all merged). Next slice: hook management LSP endpoints and handlers.


src/services/guard/GuardEngine.ts
```diff
@@ -65,6 +65,111 @@ export class GuardEngine {
...
+        const stack: string[] = [];
+        let inString: string | undefined;
+        let inRegex = false;
+        let hasContent = false;
+        for (let i = 0; i < ruleContent.length; i++) {
+            const char = ruleContent[i];
+            if (inString) {
+                if (char === '\\') {
+                    i++;
+                    continue;
```
chrisqm-dev, line 107: An if/else if block here removes the need for continue in line 107.  Same thing applies for most if statements in this method, the if chains should be simplified

src/services/guard/GuardEngine.ts
```diff
@@ -65,6 +65,111 @@ export class GuardEngine {
...
+        try {
+            const violations = this.validateTemplate(data, [rule], DiagnosticSeverity.Error);
+            return { valid: true, parseErrors: [], violations };
+        } catch (error) {
+            this.log.warn(error, 'Guard rule evaluation failed');
+            return { valid: false, parseErrors: [extractErrorMessage(error)], violations: [] };
+        }
+    }
+
+    private syntacticPreCheck(ruleContent: string): string | undefined {
```
chrisqm-dev, line 96: nit: can we rename to something that implies what it returns. Something like parseSyntaxErrors?

src/services/guard/GuardEngine.ts
```diff
@@ -65,6 +65,111 @@ export class GuardEngine {
...
+                    i++;
+                    continue;
+                }
+                if (char === '\n' || char === inString) {
+                    inString = undefined;
+                }
+                continue;
+            }
+            if (inRegex) {
+                if (char === '\\') {
```
satyakigh, line 115: Valid Guard regexes containing `/` in a character class can be rejected here. The vendored `advanced_regex_negative_lookbehind_rule.guard` uses `[A-Za-z0-9\\/+=]`; this skips the second backslash, treats `/` as the regex terminator, and later reports `]` as mismatched. Please track regex character-class/escape state and cover that fixture.

src/hooks/GuardHookPreview.ts
```diff
@@ -0,0 +1,101 @@
...
+
+        const rule: GuardRule = {
+            name: hook.typeName,
+            description: hook.typeName,
+            severity,
+            content,
+            tags: [],
+            pack: hook.typeName,
+        };
+        const violations = deps.validateTemplate(templateContent, [rule], severity);
```
satyakigh, line 80: `targetOperations` is ignored, so RESOURCE/CLOUD_CONTROL/CHANGE_SET hooks are evaluated against a raw template even though those invocations receive different inputs (`Previous`, `Changes`, per-resource data, etc.). This can produce false passes or failures. Please construct the target-specific input, or limit preview to the supported target and describe the approximation.

src/hooks/GuardHookPreview.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    try {
+        let pendingContent = ruleContentByUri.get(ruleUri);
+        if (!pendingContent) {
+            pendingContent = deps.fetchRuleContent(ruleUri);
+            ruleContentByUri.set(ruleUri, pendingContent);
+        }
+        const content = await pendingContent;
+
+        const ruleCheck = deps.validateRule(content);
```
satyakigh, line 62: This validates the rule, then rebuilds a `GuardRule` and runs `validateTemplate` again. Please centralize this in a shared `GuardEngine` evaluation API that performs the pre-check and evaluation once and returns parse errors plus violations; `validateRule` and preview can both delegate to it.

src/services/guard/GuardEngine.ts
gemammercado: implemented

src/services/guard/GuardEngine.ts
gemammercado: renamed

src/services/guard/GuardEngine.ts
gemammercado: scanner now tracks character class state, so / inside [...] doesn't end the regex. Added two tests: a char-class case and one that reads that fixture off disk

src/hooks/GuardHookPreview.ts
gemammercado: Scoped the preview to stack-target hooks rather than constructing per-target inputs here. `isPreviewableGuardHook` now requires a stack-level target — `STACK`, or empty/undefined `targetOperations` — so RESOURCE / CLOUD_CONTROL / CHANGE_SET-only hooks are excluded rather than evaluated against the wrong input shape.

The stack input itself is now built properly too: `buildStackHookInput` attaches a `HookContext`, modelled as a create (`InvocationPoint: CREATE_PRE_PROVISION`, no `Previous`, which is the documented shape for creates). Unknown fields are left as empty strings rather than invented.

Target-specific inputs need decisions I can't make correctly yet — chiefly which `InvocationPoint` to assume when no operation is in flight, and where `Previous` would come from while editing — so I've written that up as a follow-up instead of guessing. Worth noting it also has to handle intrinsic resolution: resource `ResourceProperties` arrives fully resolved, while a local template's `!Sub`/`!Ref` do not, so a rule inspecting a value produced by an intrinsic would behave differently locally. The stack target avoids this, since the stack input is the template as passed to create-stack/update-stack — unresolved, exactly as we see it on disk.

src/hooks/GuardHookPreview.ts
gemammercado: Added `GuardEngine.evaluateRule({ ruleContent, data, severity, name })`, which runs the pre-check and the evaluation once and returns `{ valid, parseErrors, violations }`. `validateRule` is now a thin wrapper over it, and the preview delegates to it — so the rule is no longer validated and then re-evaluated. The return type is exported as `RuleEvaluation` so the two call sites can't drift apart.

src/services/guard/GuardEngine.ts
gemammercado: One deviation worth flagging on this: the outer state machine is now if/else-if with zero `continue`s as you asked, but the inner per-character dispatch is still a `switch`. `unicorn/prefer-switch` fails the build if I convert that block to an else-if chain (3+ branches on the same discriminant), so the `continue`s are gone but that inner block stayed a switch.

src/services/guard/GuardEngine.ts
gemammercado: For the record I went with `findSyntaxError` rather than `parseSyntaxErrors` — it returns a single optional error (the first one found) rather than a collection, so the plural read as misleading. Happy to switch to your suggestion if you'd rather have it return all of them.

gemammercado:
Pushed an update addressing all five review comments, plus fixes from a second self-review pass. Summary of what changed since you last looked:

**From your comments**
- Regex character classes are tracked, so a `/` inside `[...]` no longer terminates the regex (@satyakigh). Covered by a test that reads the vendored `advanced_regex_negative_lookbehind_rule.guard` fixture off disk.
- Evaluation is centralized in `GuardEngine.evaluateRule`; `validateRule` and the preview both delegate to it, so nothing is evaluated twice (@satyakigh).
- `targetOperations` is honored — the preview is now explicitly scoped to stack-target hooks, and non-stack targets are excluded rather than mis-evaluated (@satyakigh).
- The pre-check's outer state machine is if/else-if with no `continue`s, and the method is renamed to `findSyntaxError` (@chrisqm-dev). Two deviations noted inline: the inner per-character dispatch stayed a `switch` because `unicorn/prefer-switch` rejects an else-if chain there, and I used `findSyntaxError` rather than `parseSyntaxErrors` since it returns a single optional error.

**Additional fixes**
- Stack input now carries a `HookContext` block (`GuardHookInput`), modelled as a create. Field casing was verified against the Guard hook application's own fixture rather than the docs, which contradict themselves on `AwsAccountId` vs `AWSAccountID`.
- A `catch` in the JSON template path swallowed its error and silently dropped the `HookContext`; it now logs before falling back. This was reachable in practice by a template mid-edit.
- Removed a `path` field from the preview violation type that only the mocks ever populated — the engine never sets it, so it was always `undefined` in production while a test asserted a value for it. cfn-guard's own message text already embeds `Path=/Resources/...`, so nothing is lost.
- Fixed a pre-check escape bug where a backslash at end of line inside a string consumed the newline, defeating the "a newline terminates a string" safeguard.
- Replaced an `as` cast with schema validation, and exported the shared evaluation result type so preview and engine cannot drift.
- Added edge-case tests for the input builder: empty template, no `Resources`, CRLF, and an existing `HookContext` on both the JSON and YAML paths.

The description is updated with the stack-target scoping rationale and the local-vs-deployed intrinsic resolution note.

gemammercado:
Two further issues raised by @satyakigh out of band, both reproduced and fixed. Also rebased onto current `main` (#677, #678).

**P1 — valid half-open range literals were rejected.** Guard supports `r[0, 65535)` and `r(0, 65535]`, but the pre-check applied normal bracket-pair rules and reported them as mismatches (the two closed forms were accepted). The scanner now recognises range literals and skips them instead of stacking their delimiters, requiring `r` at a token boundary and a same-line closer so genuine typos still fall through to normal handling. All four forms are tested, plus `r[0, 65535)]` which must still report the trailing `]`.

Worth noting this is the third false rejection in this pre-check, after the `<< >>` blocks and regex character classes — each time the scanner not knowing a Guard syntax form.

**P2 — JSON preview violations all collapsed to line 1.** The JSON path re-serialised the template with `JSON.stringify` and no indentation, flattening it before evaluation: a 17-line template became 1 line and reported its violation at line 1, column 73. `HookContext` is now injected textually right after the opening brace on the same line, so no line numbers shift — that template stays 17 lines and the violation lands at line 4, consistent with the YAML path which was already textual for the same reason.

One behavioural change falls out: when a template already declares a top-level `HookContext`, both paths now leave it as authored and log, rather than JSON overwriting and YAML appending a duplicate key whose resolution is parser-dependent (a strict YAML parser would reject the document outright). Tests pin both paths, plus JSON line preservation, `{}` injection, and a real-engine test asserting a JSON violation reports its true line.

satyakigh (approved):

chrisqm-dev (approved):

PR #681, opened 2026-09-11
feat(telemetry): add resource.type to getResource

deepfuriya:
## Summary

Adds the resource type as a low-cardinality `resource.type` dimension on the `getResource` metrics (`.count` / `.fault` / `.duration`) at both the `ResourceStateManager` and `CcapiService` scopes.

Today a `getResource` fault is emitted with the error type and error category, but not *which resource type* the call was for, so a fault metric can't tell you whether (say) an `AWS::EC2::Instance` or an `AWS::S3::Bucket` GetResource failed. This change records that type alongside the existing error dimensions, making faults diagnosable from telemetry alone.

## Changes

- **`TelemetryDecorator.ts`** — add a `resourceTypeArgIndex` option to the `Measure` / `Track` / `Count` decorators. When set, the string argument at that index is attached as the `resource.type` metric dimension. It's applied after context attribute extraction and guarded on a non-empty string, so it's a no-op for missing/non-string arguments and doesn't affect existing behavior.
- **`ResourceStateManager.getResource`** and **`CcapiService.getResource`** — opt in via `resourceTypeArgIndex: 0` (their first argument is `typeName`).

## Design notes

- The primary **identifier** (the resource's ID/name) is intentionally **not** added. It is high-cardinality and can contain user data, so emitting it as a metric dimension would both explode the metric series and risk leaking resource names into telemetry. Only the bounded resource type is recorded — consistent with the existing `resource.type` dimension already used for completion/hover.

## Testing

- Added unit tests in `tst/unit/telemetry/TelemetryDecorator.test.ts` covering:
  - `resource.type` attached from the configured argument index (sync and async)
  - no `resource.type` attached when the argument is missing or not a string
- Existing decorator tests (including context-attribute extraction) remain unchanged and continue to pass.


src/telemetry/TelemetryDecorator.ts
```diff
@@ -8,6 +8,8 @@ type ScopeDecoratorOptions = {
 type ScopedMetricsDecoratorOptions = {
     name: string;
     extractContextAttributes?: boolean;
+    /** Attach the string argument at this position as the `resource.type` metric dimension. */
+    resourceTypeArgIndex?: number;
```
Zee2413, line 12: Ideally telemetry decorator does not contain specific business logic based options. Can we emit the resource type details via an attribute from within the service/manager instead without expanding the decorator?

satyakigh, line 12: We already have the args available in the decorator, why do we need a new field?

deepfuriya (commented):
Discarding this PR since no longer required

